Issue #10, opened 2017-02-08
Missing standard Command Line Interface in OpenSDS


leonwanghui:
Currently OpenSDS CLI is just a simple parser with string. To make it more convenient to use, we should develop OpenSDS CLI as a well-accepted function, just like OpenStack and Kubernetes (kubectl).

leonwanghui:
This issue is solved.

PR #11, opened 2017-02-08
enhance OpenSDS CLI function and add logging module

leonwanghui:
To solve Issues #10, after taking a reference with Kubernetes and Torus (by CoreOS), we found that Cobra, a library for creating powerful modern CLI applications, can help us enhance OpenSDS CLI function. So we did some change to the code to make CLI a better user experience. Besides, we add logging module in to debug the system.


leonwanghui:
Apologize for the wrong word "CNI", which is actually "CLI".

PR #14, opened 2017-02-09
reorginize the structure of code to pass GitHub CI

leonwanghui:
In this update, we made modifies with the whole project structure to make it an awsome project.


leonwanghui:
Abandon this PR and open another one.

PR #28, opened 2017-04-10
Opensds external provisioner design

leonwanghui:
In this update, we launched a proposal about OpenSDS external provisioner, to provide a unified provisioner for different backend storage types.


leonwanghui:
Closed

PR #33, opened 2017-04-13
OpenSDS Southbound Interface Refactor

leonwanghui:
In this patch, we drafted a proposal about refactoring OpenSDS southbound interface, to make it more easy for backends to implement.


leonwanghui:
Closed

PR #38, opened 2017-08-06
Return it when error occurs

leonwanghui:
In dock api module, if some errors occur during deleting volume
attachment, the error should be returned rather than being ignored.
So it was a bug and should be fixed immediately.


leonwanghui:
Since nobody has question, this PR is ok to be merged.

PR #39, opened 2017-08-07
Move southbound drivers into contrib directory

leonwanghui:
Since OpenSDS Controller project only covers the resource management
and orchestration, it would be confusing if some southbound drivers
lie in dock module of this project. So we plan to move all southbound
drivers into contrib directory in the first stage, and finally we
suggest removing all of them and forming a new project that only
contains southbound storage drivers.


leonwanghui:
This PR will be rebased, so close it now.

PR #42, opened 2017-08-08
Refactor dock server structure in grpc module

leonwanghui:
In the go formal way, all structure constructing function should
return with interface, one of reason is that they need to comply
with the inheritance conduction of unified interface. Besides,
the seperation between structure and interface will make it easier
for unit test.


leonwanghui:
Looks like nobody has comments, I will merge this PR at first and develop more test codes in this package.

PR #53, opened 2017-08-15
Add the api design of sdscontroller v1alpha1 version

wisererik:
According to the project [roadmap](https://github.com/opensds/opensds/wiki/OpenSDS-Controller-Project-Roadmap), there are some feature will be completed in the fourth quater of 2017. Therefore, the api about these feature should be designed in advance and  it's alpha version at present.


leonwanghui:
@wisererik Hi Erik, can you add some description about this spec, such like the motivation or some plans? Thanks!

wisererik:
@leonwanghui thanks for reminding, the comments has been added :)

leonwanghui:
@wisererik Everything looks great, but only one question:
If I get it right, for backend, profile and pool these three models, the current design is to devide them with storage types (block, file and object). But IMO this may cause some troubles in some use cases.
So I will merge it and open another issue to discuss some hidden troubels. Is that ok?

PR #57, opened 2017-08-24
Controller Framework Refactor

leonwanghui:
## Motivation
To solve some problems of testing work as a reference to [issue 52](https://github.com/opensds/opensds/issues/52), we did a lot of work to refactor the current framework of controller module and make up the testing code.
## Changes
All the changes begin from testing. To make the code easier for testing, we added a lot of interfaces for mocking the data in some use cases.
## Next-step
- More testing code
- Redesign a stronger framework for better scaling out
- Fix some bugs

PR #58, opened 2017-08-26
Fix some remaining issues

leonwanghui:
## Motivation
As a reference to [# 57](https://github.com/opensds/opensds/pull/57), this PR is aiming at making up some blanks of testing work and fixing some remaining bugs, which is our long-term goal.
## Changes
Changes of this PR are more than belows:
- Move some public methods to ```util``` package
- Add a ```Setter``` interface for testing
- Remove ```fake.go``` file in ```client``` package and move all stuffs to ```volumecontroller_test.go```
- Redefine the meaning ```Endpoint``` field from ```host_ip``` to ```host_ip:port```
- Add some testing files (finished and unfinished)

That's all, any thoughts?


leonwanghui:
Since there is no comment during these days, I will merge this PR right now.

More bug reports and fixed patchs are welcomed : )

PR #63, opened 2017-09-04
Manage opensds dependency packages through dep tool

leonwanghui:
## Motivation
As we know, opensds controller project is light-weighted because it has no external-dependencies, which means opensds needs to manage all dependencies itself. But for now this project didn't do anything to manage these packages.
## Solution
To solve this problem, I found a popular Go dependency management tool called [dep](https://github.com/golang/dep), and after some experiments I think it can reduce a lot of complicated work for managing dependency packages through this tool.

Here are some outputs when using this tool:
```
krej@krej-Lenovo-IdeaPad-Y470:~/gopath/src/github.com/opensds/opensds$ dep status
PROJECT                            CONSTRAINT     VERSION        REVISION  LATEST   PKGS USED
github.com/asaskevich/govalidator  *              v6             4918b99   4918b99  1   
github.com/astaxie/beego           ^1.9.0         v1.9.0         d96289a   d96289a  9   
github.com/bouk/monkey             branch master  branch master  b96e337   b96e337  1   
github.com/ceph/go-ceph            branch master  branch master  81e4191   81e4191  2   
github.com/coreos/etcd             ^3.2.7         v3.2.7         bb66589   bb66589  5   
github.com/go-openapi/errors       branch master  branch master  03cfca6   03cfca6  1   
github.com/go-openapi/runtime      branch master  branch master  bf2ff8f   bf2ff8f  2   
github.com/go-openapi/strfmt       branch master  branch master  610b6ca   610b6ca  1   
github.com/go-openapi/swag         branch master  branch master  f3f9494   f3f9494  1   
github.com/go-sql-driver/mysql     ^1.3.0         v1.3           a0583e0   a0583e0  1   
github.com/golang/protobuf         branch master  branch master  17ce142   17ce142  5   
github.com/mailru/easyjson         *              branch master  2a92e67   2a92e67  3   
github.com/mitchellh/mapstructure  *              branch master  d0303fe   d0303fe  1   
github.com/satori/go.uuid          ^1.1.0         v1.1.0         879c588   879c588  1   
golang.org/x/net                   branch master  branch master  66aacef   66aacef  8   
golang.org/x/text                  *              branch master  bd91bbf   bd91bbf  9   
google.golang.org/genproto         *              branch master  ee236bd   ee236bd  1   
google.golang.org/grpc             ^1.6.0         v1.6.0         f92cdcd   f92cdcd  15  
gopkg.in/mgo.v2                    *              branch v2      3f83fa5   3f83fa5  2   
[...]

PR #64, opened 2017-09-05
Add storage pool feature to ceph

xxwjj:
Due to the ceph plugin do not support StoragePool feature , we created ceph image use default pool(rbd) before.

So I implment it.


leonwanghui:
@xxwjj  Hi Jerry, glad to see you have implemented it, but there is something confusing in this patch:
1. According to the ```StoragePoolSpec``` defined in opensds api, I think you missed a field called ```parameter```  which contains all storage features (such as replica number, IOPS and so on) of this pool. And IMO it plays a very important role in storage discovery process, because those features are basics of policy-based orchestration function.
2. After reading your code, I have been wondering if it's better for every function to return a pointer to that structure. Because I found that every structure returned contains a lot of field and it could waste a lot of spaces.

Just my 2 cents :p

xxwjj:
@leonwanghui hi,leonwanghui,great suggestions. But there is problem i think we need to disscuss on next weekly meeting.Because of the cpeh is  a unified, distributed storage system ,some attributes are different from traditional storage pools. such as IOPS,bandwith etc. and these paramter is very important for scheduling. You can see the detail information 
 about the ceph storage pool [here](http://docs.ceph.com/docs/master/rados/operations/pools/)

leonwanghui:
@xxwjj Sorry for replying to you up to now. I noticed that there is something wrong with OpenSDS package dependencies currently, and I have pushed a patch [#64](https://github.com/opensds/opensds/pull/63) to solve this problem. Anyway, what we expect is to find a stable and extensible way to avoid all hidden dangers, rather than fixing bugs after they occur. So I will attach this topic to the next meeting and we could have a disccusion then.

BTW, from your latest patch, there is some exec command (such as ```execCmd("ceph df")```) which could need some packages before building work. So can you have a check and make sure if there are some missing configuration at ```travis.yml``` file?

xxwjj:
@leonwanghui 

    yeah, I agree with you, we should discuess it on next weekly meeting, and  find the root cause.But there is a [issue ](https://github.com/coreos/etcd/issues/8524)in etcd may be some reason for it.

    Due to the ceph api interface provided by libraods librbd is so poor, so I can only get more information by ceph command.The travis CI is just for building and Unite test right now, so there is nothing we need add to travis.

leonwanghui:
Since this patch has conflict with PR [#68](https://github.com/opensds/opensds/pull/68), it will be closed now.

PR #65, opened 2017-09-18
Add global configuration  utils

xxwjj:
Add global configuration utils. The configuration data is read from the user specified configuration file and initialized to the corresponding variables.So user can simplely use configuration like this:

```	
        utils.CONF.Load("/etc/opensds/opensds.conf")
        ...
	ds := dockServer.NewDockServer(utils.CONF.OsdsDock.ApiEndpoint)
```


pkg/utils/config.go
```diff
@@ -0,0 +1,100 @@
+package utils
+
+import (
+	"github.com/Unknwon/goconfig"
```
wisererik: This library is not maintained, you can consider the other library, such as [ini](https://github.com/go-ini/ini)

pkg/utils/config.go
```diff
@@ -0,0 +1,100 @@
...
+		default:
+		}
+	}
+}
+
+func initConf(confFile string, conf interface{}) {
+	cfg, err := goconfig.LoadConfigFile(confFile)
+	if err != nil {
+		log.Println("[Info] Read configuration failed, use default value")
+	}
```
wisererik: When the osdsdock or osdslet daemon is started, this configuration will be initialized, so where will the log be saved?

pkg/utils/config.go
xxwjj: @wisererik good suggestion, the main authors of both libraries are Unknown, there wold be many similarity between the two libraries. and it seems that [INI](https://github.com/go-ini/ini) community is more active and more popular.

pkg/utils/config.go
xxwjj: @wisererik yeah, I considered this question too, at the time. Maybe we can use fmt.Printfln to instead. Do you have any better ideas.

leonwanghui:
Looks good to me :b

PR #68, opened 2017-09-19
Add ceph list pool feature

xxwjj:
Add ceph list pool feature, the origin branch is confict, so create new branch. You can see [here ](https://github.com/opensds/opensds/pull/64)for more details .

### DIFF
--- a/contrib/plugins/ceph/ceph.go
+++ b/contrib/plugins/ceph/ceph.go
@@ -29,12 +29,36 @@ import (
 	"github.com/ceph/go-ceph/rados"
 	"github.com/ceph/go-ceph/rbd"
 	"github.com/satori/go.uuid"
+	"os/exec"
+	"strconv"
 )
 
 const (
 	OPENSDS_PREFIX string = "OPENSDS"
 	SPLIT_CHAR            = ":"
 	SIZE_SHIFT_BIT        = 20
+)
+
+const (
+	GLOBAL_SIZE = iota
+	GLOBAL_AVAIL
+	GLOBAL_RAW_USED
+	GLOBAL_RAW_USED_PERCENTAGE
+)
+
+const (
+	POOL_NAME = iota
+	POOL_ID
+	POOL_USED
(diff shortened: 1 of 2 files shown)


contrib/plugins/ceph/ceph.go
```diff
@@ -356,6 +391,136 @@ func (imgMgr *ImageMgr) GetSnapshots() (*[]SnapshotResponse, error) {
...
+	var pools []PoolResponse
+	for i := range pc {
+		name := pc[i][POOL_NAME]
+		param := imgMgr.getPoolParam(pa[name])
+		totalCap := imgMgr.parseCapStr(gc[GLOBAL_SIZE])
+		maxAvailCap := imgMgr.parseCapStr(pc[i][POOL_MAX_AVAIL])
+		availCap := imgMgr.parseCapStr(gc[GLOBAL_AVAIL])
+		pool := PoolResponse{
+			Name:       name,
+			ID:         uuid.NewV5(uuid.NamespaceOID, name).String(),
```
leonwanghui: Looks great, but one thing need to be careful here is we used to uuid.NewV4() to create uuid, why using NewV5() here? Could it cause some hidden bugs?

xxwjj: @leonwanghui The UUIDV5 is not upgrade version of UUID V4. Their main difference is the implementation of the algorithm. Version 5  UUID use the SHA1 hash functions , to combine a namespace with a piece of already unique data to generate a UUID. Version 4 UUIDs essentially use just 16 bytes of randomness  to generate a UUID.More detail infomation you can look at [here](http://stackoverflow.com/questions/703035/when-are-you-truly-forced-to-use-uuid-as-part-of-the-design/786541#786541). I use UUIDv5 just because I want to use ceph pool name to  generate the UUIDs so that UUID will be a fix value.

leonwanghui: ok, great :p

PR #73, opened 2017-09-21
move files into dock package from the grpc package

wisererik:
### Introduction

IMO, grpc directory files such as client, server and protobuff definition should belong to the dock service to achieve the framework part, so it's better to move them into the dock directory.

### Changes

- Move files into dock package from the grpc package.


leonwanghui:
Sounds good to me :b

Issue #46, opened 2017-08-09
Add log level features to opensds


xxwjj:
OpenSDS log system is so poor,we just using printing string to diff log level. 

We could have referenced k8s's way and implemented it using  [glog](https://github.com/golang/glog).

leonwanghui:
Yes you are right, that is definitely something we should do. Besides glog, I think [logrus](https://github.com/sirupsen/logrus) is also a good choice.

BTW, @xxwjj can you add it on our weekly meeting [agenda](https://docs.google.com/document/d/1OKDp7iDw_l_5kGyniqTKtrwRFAZMvItGghsxfrCqr5c/edit#) so that we can discuss it on the next weekly meeting?

xxwjj:
ok

leonwanghui:
This Issue will be closed, for its referenced [PR](https://github.com/opensds/opensds/pull/74) has been merged.

PR #74, opened 2017-09-22
Add log level features to opensds

xxwjj:
This PR is for Issue #46 . Two things should be notified.

1、The new log system is implemented by glog. so that we should import 'github.com/golang/glog' for develpment later.
2、The current code is not affected, and these log infomations will be printed to the same log file.

### DIFF
--- a/cmd/osdsdock/osdsdock.go
+++ b/cmd/osdsdock/osdsdock.go
@@ -20,15 +20,12 @@ This module implements a entry into the OpenSDS REST service.
 package main
 
 import (
-	"fmt"
-	"log"
-	"os"
-	"strings"
-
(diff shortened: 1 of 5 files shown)


xxwjj:
@leonwanghui This PR just provide framwork, and is compatible with previous log system. The usage of log level would be like this.

 ```
package main

import (

	"github.com/golang/glog"
)

func main() {
	glog.Info("Info test")
	glog.Warning("Warning test")
	glog.Error("Error test")
	glog.Fatal("Fatal test")
	glog.Flush()
}
```

The log code we have writted before should be modified manuly, and I will do it in next PR.

leonwanghui:
Ok

Jerry <notifications@github.com>于2017年9月22日 周五14:22写道：

PR #75, opened 2017-09-22
Modified global config directory structrue

xxwjj:
Modified global config directory structrue，make is easy to understand.


leonwanghui:
+1

PR #78, opened 2017-09-23
Update dock module

leonwanghui:
## Introduction
In this update, there are two major changes:
- Update discovery operation in dock module. I created an abstract interface ```Discoverer``` which contains several methods. The ```Init``` method indicates the initialization of DockDiscoverer structure and it will discover all dock resources. The ```Discovery``` method indicates how to discover storage pool resources from all dock saved in DockDiscoverer, and after it done the ```Store``` method will store all information (dock, pool) in database.
- Change the ```NewSetter``` initializing work by creating a global variable called ```S``` in its package. IMO in this way it can reduce some costs when calling some private method of this structure.

Besides, considering some of files are created in this year (2017), I updated the license of related files.


wisererik:
@leonwanghui The `Discovery` interface seems to have nothing to do with the dock server that located on the package /pkg/dock/server, I guess you want to find the dock, the pools and store them in the database. IMO, the main process is done when the dock server is registered:

1. Dock server read the dock config file and initialize the dock server, such as port and construct `DockSpec` object.
2. Dock server read the driver config file and initialize the driver and relate it to dock server.
3. Dock server discovery all pools' information configured in driver config.
4. Dock server store dock information and pools'information in the database when the above operation is successful.
5. Start the dock server(start gRPC service).

leonwanghui:
@wisererik Thanks for your suggestion. First I will give a simple description about the Discoverer here.
```Discoverer``` interface is designed for exposing the feature of ```DockDiscoverer``` structure to outside world. As you can see from ```Discovery``` method below, I think it acts the similar function compared to your solution.
```go
func Discovery(d Discoverer) error {  
	var err error  
  
	if err = d.Init(); err != nil {  
		return err  
	}  
	if err = d.Discovery(); err != nil {  
		return err  
	}  
	if err = d.Store(); err != nil {  
		return err  
	}

  	return err  
  }  
```
But one thing to be done is config dock resources directly from global config structure, rather than reading from a standalone config file. I will ask Jerry for fixing it.

PR #81, opened 2017-09-24
[WIP] Southbound interface design through protobuf

leonwanghui:
Hi all, I designed the first version of southbound interface in this update.


wisererik:
@leonwanghui I don't agree this pr, plz explain your design ideas.

leonwanghui:
@wisererik Hi, I will try to solve your doubts with some explainations about this patch below:

As we know, the essential goal of designing southbound interface is to provide a unified storage resource management interface so that multiple drivers can easily be integrated into OpenSDS. To simplify the development work, I chose protobuf to design all models.
As you can see from ```dock.proto``` file, I used some of existing models (such like ***Opts) as input parameters of every method, and I also added some models so that all drivers only need to comply with grpc model. For more information you can refer to the modified ```sample.go``` file.
If this change is agreed, then in ```dock.go``` file these models defined in protobuf should be converted to models defined in ```model``` module and then saved into database.

If you still have questions, plz point them out and I will slove it as far as I could.

wisererik:
@leonwanghui IMO, the model.StoragePoolSpec has the same ability as pb.StoragePool, the sample in sample.go can be represented by model.StoragePoolSpec. So, I don't understand why it's have to be redefined.

leonwanghui:
I can explain it with these two reasons:
1. Firstly, the reason why I created another structure in proto model is that in this way vendor drivers only need to import one package.
2. Next, as you can see in pb.StoragePool, its scemantics is much like Swordfish which is a storage management standard model defined by SNIA. What I am trying to do is to adopt Swordfish as our southbound interface.

Does it make clear to you?

leonwanghui:
Since bugs are all gone, the patch will be merged.

PR #82, opened 2017-09-25
Add ceph config and fix some typo and bugs

xxwjj:
1、Add ceph config feature.
2、Fix global config omit bugs.
3、Modified const typo in ceph.


leonwanghui:
@xxwjj  From the error of building work, I guess you passed a ceph config file ```/etc/opensds/ceph.yaml``` to initialize ceph config when running unit test, which result in the error that it couldn't find the config file. I suggest you pass a whole fake ```CephConfig``` structure into it so that it could pass the testing work.

leonwanghui:
@xxwjj I have a suggestion that you can return nil when error occurs.

leonwanghui:
+1

PR #83, opened 2017-09-27
Fix some some bugs which relative to ceph driver and add some missing function.

xxwjj:
Fix blow bugs.
1、BandWidth typo error.
2、Change yaml package to gopkg.in/yaml.v2
3、Add ceph command specified the config file function.
4、Change some log printing format.


leonwanghui:
Much better now, thans jerry

PR #85, opened 2017-09-28
Implement the discovery feature

xxwjj:
Implemet the discovery feature.User should modified two configuration file.
1、Modified global configuration file,the default file path is `/etc/opensds/opensds.conf`. Add backend driver names to the key  `enabled_backends` to enable the relative backend drivers, the driver names should be separate by comma. Then add driver basic info section. Possible example maybe like blow:
```
#...
[osdsdock]
enabled_backends = ceph
ceph_config = /etc/opensds/driver/ceph.yaml
#...
[ceph]
name = CephDriver
description = Ceph Test
endpoint = 127.0.0.1
driver_name = ceph
```
2、For some drivers, you need to configure the specified configuration file(say ceph driver the default file path is `/etc/opensds/driver/ceph.yaml`) for more detailed configuration.Possible ceph example maybe like blow:
```
configFile: /etc/ceph/ceph.conf
pool:
  "rbd":
    diskType: SSD
    iops: 1000
    bandWitdh: 1G
  "test":
    diskType: SAS
    iops: 800
    bandWitdh: 800M
```
In this file we should config the pool that Opensds can use, and some detail paramters diskType,iops,bandWitdh.


leonwanghui:
+1

PR #88, opened 2017-09-30
Modified the ceph driver to adapt the southbound changes

xxwjj:
Due to the OpenSDS southbound  has changed, so ceph driver need to modify .

### DIFF
--- a/contrib/drivers/ceph/ceph.go
+++ b/contrib/drivers/ceph/ceph.go
@@ -156,9 +156,9 @@ type Driver struct {
 	ioctx *rados.IOContext
 }
 
-func (d *Driver) Setup() {}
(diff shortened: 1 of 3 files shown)


leonwanghui:
@xxwjj It looks good, but IMO I prefer using ```GetId()``` rather than ```Id``` when fetching a filed in structure:
```go
name := opt.Name
size := opt.Size
```
Because usually we use ```opt.Name``` to change Name field value of opt, what do you think?

xxwjj:
@leonwanghui ok, I will change it.  but I am afraid this usage design is too complex.

leonwanghui:
I think it's a good way to prepare for designing some read-only field, because in this way we just need to change the first letter of field from upper to lower, rather than modifying too much code.

PR #91, opened 2017-10-11
Add sample driver for test

xxwjj:
Modifed config and discovery to support the sample driver, so that developer can just modify the config file to used sample driver for testing.


leonwanghui:
+1

PR #94, opened 2017-10-12
Fix bug when finding version by specified version name

wisererik:
### Changes

- Fix bug when finding version by specified version name.

- Add unit test for version.go.


wisererik:
@leonwanghui thanks for your suggestion, it's has been modified.

PR #98, opened 2017-10-19
Change cinder config file format

xxwjj:
The format of cinder configuration file is json, and it's little complex, so I change it to yaml format.

So the cinder configuation file would be like this:

``` yaml
authOptions:
  endpoint: "http://192.168.56.104/identity"
  domainId: "Default"
  domainName: "Default"
  username: "admin"
  password: "admin"
  tenantId: "04154b841eb644a3947506c54fa73c76"
  tenantName: "admin"
```


leonwanghui:
Nice change!

PR #100, opened 2017-10-20
Add pool config feature

xxwjj:
Add pool config feature. Adminitrator can add some properties(DiskType, IOPS, BandWidth etc.) to the pools so that OpenSDS scheduler can alloc volume according these properties automatic.And it will also takes on the white list function, that the the user only can see the pool which is added into the config file.The configuration would be like this:
``` yaml
pool:
  pool1:
    diskType: SSD
    iops: 1000
    bandwidth: 1000
  pool2:
    diskType: SAS
    iops: 800
    bandwidth: 800
```


leonwanghui:
Cool!

PR #101, opened 2017-10-20
[WIP] Add integration test

leonwanghui:


.travis.yml
```diff
@@ -19,6 +22,7 @@ env:
 before_install:
   - sudo apt-get update
   - sudo apt-get install librados-dev librbd-dev -y 
+  - sudo docker run -d --net=host leonwanghui/osdsdock:integration-test
```
hannibalhuang: leonwanghui/osdsdock ? should it be the opensds repo ?

leonwanghui: @hannibalhuang Yeah it should be, but since we haven't registered docker account of opensds team, it's illegal to push into opensds repo. But it's indeed a problem which should be settled down recently.

hannibalhuang: okey understood

Issue #102, opened 2017-10-21
Add client package for opensds controller project


leonwanghui:
## Motivation
As a collaborative project, opensds controller project will work with other projects such as OpenStack, Kubernetes and so forth. To better integrated with ```nbp``` and other Golang project, it's necessary to design client package for opensds controller project to set up connection with opensds service.

## Introduction
Since we have adopted ```beego``` as the framework of api-server, I suggest we also use it to implement the http client module. Besides, we can utilize all global module (model, config, log etc) to have minimum workload.

Any thoughts?

leonwanghui:
Resolved.

PR #104, opened 2017-10-23
[WIP] Add client package for opensds controller project

leonwanghui:
This patch targets as issue [#102](https://github.com/opensds/opensds/issues/102), more info can be found in README.md.


wisererik:
@leonwanghui Nice job, so the [nbp](https://github.com/opensds/nbp) can use this library~

client/dock.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if err := d.Recv(request, url, "GET", d.Opt, &res); err != nil {
+		fmt.Println(err)
+		return nil, err
+	}
+
+	return res, nil
+}
+
+func (d *DockMgr) ResetAndUpdateDockRequestContent(in interface{}) error {
```
xxwjj: This method name  is too loog, it`s not good for reading, can you make a shorter one?

leonwanghui: Ok, I'll have a try, thanks jerry!

xxwjj:
@leonwanghui There are too points I think we need disscuss . 
1、One developer use the client funtcion  invoke tow method, IMO, it's too cumbersome to use.The fllowing usage maybe better:
```golang
package main

import (
	"fmt"
	
	"github.com/opensds/opensds/client"
)

func main() {
	c := NewClient(&client.Config{
		Endpoint: ":8080",
	})
	
	fmt.Printf("Updated c is %v\n", &modelVolumeSpec{Name: "test"}))
	vol, err := c.CreateVolume()
	if err != nil {
		panic(err)
	}
}
```
2、For some simple scenarios, the parameters can not be encapsulated.For example the DeleteVolume funtcion, we just only send a volume of ID to it.

leonwanghui:
@xxwjj Thanks for your suggestion, I have updated the patch and plz have a look.

PR #105, opened 2017-10-23
Add unit test for dock api

wisererik:
### Changes

- Add unit test for dock api.
- Change the prefix of the url in router.
- Mock database interface by testify/mock library, the package mock provides a system by which it is possible to mock your objects and verify calls are happening as expected. see details about how to use [testify/mock](https://github.com/stretchr/testify).


leonwanghui:
@wisererik From the log error printed, I guess its cause is the missing import of ```testify/mock``` package, so I suggest you use glide tool to pull it locally by running ```glide get github.com/stretchr/testify``` command.

Hope it could offer some help : )

pkg/api/dock.go
```diff
@@ -59,17 +59,21 @@ func (this *DockPortal) ListDocks() {
...
-		reason := fmt.Sprintf("Get docks failed: %s", err.Error())
-		this.Ctx.Output.SetStatus(StatusBadRequest)
+		reason := fmt.Sprintf("Get dock failed: %s", err.Error())
+		this.Ctx.Output.SetStatus(StatusInternalServerError)
+		this.Ctx.Output.Body(utils.ErrorStatus(this.Ctx.Output.Status, reason))
+		log.Error(reason)
+		return
+	}
+
+	if result == nil {
```
leonwanghui: Honestly, I don't think it's neccesary to add this sentence, for the reason that every time the result is nil, the err must be ```not nil```, so it can be assured that the result is ```not nil``` by asserting if err is equal to nil.

wisererik:
@leonwanghui thanks for your suggestion, plz check again~

leonwanghui:
Cool, I will merge this patch first, so can you fix the missing part of api module later?

PR #108, opened 2017-10-24
Add unit test for pool api and fix some bugs in router

wisererik:
### Changes

- Add unit test for pool api
- Fix bugs in router about the prefix of url


leonwanghui:
Sounds good.

PR #113, opened 2017-10-26
Update delete request in client package

leonwanghui:
In this update, to improve user-experience, the ```Response``` value was removed when user sends a delete request.


wisererik:
+1

PR #115, opened 2017-10-27
add unit test for volume

wisererik:
### Changes

- Add unit test for volume.
- Add unit test for volume snapshot.


leonwanghui:
Good job!

Issue #116, opened 2017-10-30
Add LVM Driver Support


leonwanghui:
## Motivation
As a storage management and orchestration layer in multi-cloud environment, it's vital for opensds controller project to enlarge its southbound eco-system. Currently we have Ceph and Cinder driver which stand for storage system, but we still miss something in storage device management.

## Goal
To solve this problem, we are trying to add LVM storage device driver, which can be considered as a starting point. The reason for pushing this suggestion out is that we can prove it that eventually opensds will become a storage controller that manages different kind of storage resources.

## Plan
IMO, this design should be finished before alpha release, so I think we should start it ASAP if nobody has objection.

leonwanghui:
Closed.

PR #118, opened 2017-11-01
[WIP] Add LVM driver support

leonwanghui:
In this patch, I added the partial implementation of lvm driver (volume CRUD, snapshot CRUD). And to compatible with current framework, I did a bit change of OpenSDS model, which is adding ```Metadata``` field in VolumeSpec and VolumeSnapshotSpec.


leonwanghui:
Add a reference with Issue [#116](https://github.com/opensds/opensds/issues/116)

contrib/drivers/lvm/lvm.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+type LVMConfig struct {
+	Pool map[string]PoolProperties `yaml:"pool,flow"`
+}
+
+type PoolProperties struct {
+	DiskType  string `yaml:"diskType"`
+	IOPS      int64  `yaml:"iops"`
+	BandWidth int64  `yaml:"bandwidth"`
```
wisererik: BandWidth or Bandwidth?

wisererik:
Hi, Leon, can the binary files in the integration test directory be automatically generated when the test is excuted?

contrib/drivers/lvm/lvm.go
leonwanghui: It should be Bandwidth, but I didn't have so much energy to solve it because it's a bit complicated.

leonwanghui:
@wisererik Yeah, you can modify it in travis.yml file.

PR #123, opened 2017-11-06
Update controller initializing operation

leonwanghui:
As talked before, it's complex to test if we put all ***Opts struct in Controller struct, so in this patch I transfer these fields to request parameters.


wisererik:
@leonwanghui nice job!

PR #124, opened 2017-11-07
Add update volume attachment interface

wisererik:
### Changes

1. Add attachment update interface.
2. Update the DB interface of attachment.

### DIFF
--- a/client/volume.go
+++ b/client/volume.go
@@ -102,6 +102,18 @@ func (v *VolumeMgr) CreateVolumeAttachment(body VolumeAttachmentBuilder) (*model
 	return &res, nil
 }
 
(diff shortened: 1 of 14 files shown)


leonwanghui:
Hi @wisererik, I suggest you use ```if atc.Id == ""``` rather than ```if len(atc.Id) == 0``` when asserting a variable. Besides, it's still something wrong with CI work.

wisererik:
@leonwanghui thanks for your suggestion, both `len(str) > 0` and `str != ""`  are clear, efficient. Besides, I found the integration test failed in CI job, but I didn't find the problem, could you help me find where is the CI error?

leonwanghui:
Good job!

PR #125, opened 2017-11-08
Improve some outputs of opensds client when failing to send a request

leonwanghui:


leonwanghui:
Since this patch didn't update some methods in client module, which should be done, there is no impact for opensds upstream projects.

PR #127, opened 2017-11-10
Add CSI driver to simplify storage orchestration for cloud-native apps

leonwanghui:
## Background

The general goal of CSI is to define an industry standard “Container Storage Interface” that will enable storage vendors (SP) to develop a plugin once and have it work across a number of container orchestration (CO) systems.

## Motivation

Currently it's difficult for CO to find where the volume is distributed in multi-plugins co-existing scenario. For example, there exist two CSI plugins (Ceph and Cinder) in two nodes, firstly CO create a volume in Ceph Controller Plugin, then when CO wants to publish this volume, it has no idea which plugin should call because itself doesn't perceive plugin info.

The second problem we met is that there is not so many storage advanced features (snapshot, backup, replication, etc) in CO, which results in the inconsistency of capability between CO and storage systems.

## Solution
After some discussion, we find that OpenSDS is the best solution for solving all problems above, and ```most importantly``` it doesn't change a line of code in current COs.
For more information, please refer to: https://docs.google.com/presentation/d/1wn0msm129-kVrTc-890kCmlpPbAm-qNgXmX_EZ8HjQI/edit?usp=sharing

More suggestions are welcomed : )


leonwanghui:
I will move the deadline of this patch  to next release, because we have to publish the release recently.

xing-yang:
What's the status of this PR?

leonwanghui:
Currently it's pending.

skdwriting:
@xing-yang @leonwanghui this looks to be a very old PR. Can we close this PR?

xing-yang:
@leonwanghui I'm going to close this PR now as it has been stale for a year.  Please re-open if you need to work on it again.  Thanks.

PR #128, opened 2017-11-14
Optimize the global configuration usage for driver developer.

xxwjj:
Packaged the storage backends in the structure `Backends`. So one storage backend driver  can be added by modifying only one part of configuation structure code.


leonwanghui:
Looks great!

PR #129, opened 2017-11-15
Optimized the configuration

xxwjj:
Modified the driver relative configuration method and Optimized the global configuration.


leonwanghui:
Since there is something wrong with travis itself not the code, this patch will be merged.

PR #130, opened 2017-11-17
Update DockHub initialization and improve discovery operation

leonwanghui:
In this patch, I changed DockHub to become global struct, and moved DockDiscoverer struct into DockHub to make it more clear about relationship between them.


leonwanghui:
This patch need be merged due to integration test.

PR #144, opened 2017-11-22
Fixed some bugs

xxwjj:
Fixed some bugs blows:
1. CLI volume and pool add more items to display.
2. Fixed ceph command parsing bugs.
3. Modifed  some logs so that developers can get more usefull infomations.


leonwanghui:
Good job!

PR #149, opened 2017-11-23
Fix some lvm bugs

xxwjj:
The lvm driver does not support the mult-VolumeGroups scenario in current version, so I submit this PR to fix it.


leonwanghui:
@xxwjj Well done!

PR #160, opened 2017-12-05
Add global package for generating url.

leonwanghui:
In this patch, I tried to decouple some descriptive fields in current model into a single spec called ```Descriptor```, we could benefit from it to classify fields for different usage.


leonwanghui:
This patch was to modify opensds model, but it seems that a new patch is merged into it so I update this patch to avoid some confusions.

PR #173, opened 2017-12-11
Simplify the Set/Get method of the model.

wisererik:


leonwanghui:
Hi, it seems that the make work doesn't work, plz make sure it running locally before pushing it.

wisererik:
@leonwanghui tks, I have resolved it.

leonwanghui:
Great

PR #176, opened 2017-12-12
Add unit test for dock cli

PengYiOfOpenSDS:


leonwanghui:
@PengYiOfOpenSDS  It seems to have some build error according to the CI result, plz have a check.

wisererik:
@PengYiOfOpenSDS the pkg/cli has been moved into parent package, the same as pkg.

PR #192, opened 2017-12-18
Add unit test for api

PengYiOfOpenSDS:


pkg/api/volume_test.go
```diff
@@ -216,6 +216,41 @@ func TestUpdateVolume(t *testing.T) {
 	}
 }
 
+func TestUpdateVolumeWithBadRequest(t *testing.T) {
+	var jsonStr = []byte(``)
+	r, _ := http.NewRequest("PUT",
+		"/v1beta/block/volumes/f4a5e666-c669-4c64-a2a1-8f9ecd560c78", bytes.NewBuffer(jsonStr))
```
leonwanghui: I suggest you import ```github.com/opensds/opensds/pkg/utils/urls``` to generate url automatically so that we don't nee d to change it anymore.

PR #194, opened 2017-12-18
Add no authentication options to access cinder without authentication.

wisererik:
When cinder is configured as `noauth` mode, the keystone authentication is not requred, and the current cinder driver in opensds does not support this. Therefore, I added some configure options to support the `noauth` mode,


leonwanghui:
@wisererik  Hi, can you give a simple description about your patch? Thx!

PR #195, opened 2017-12-19
Add functions UpdateVolume and UpdateVolumeSnapshot for the client

PengYiOfOpenSDS:


client/volume_test.go
```diff
@@ -196,6 +196,37 @@ func TestDeleteVolume(t *testing.T) {
...
 }
 
+func TestUpdateVolume(t *testing.T) {
+	var volID = "bd5b12a8-a101-11e7-941e-d77981b584d8"
+	vol := model.VolumeSpec{
+		Name:        "sample-volume",
+		Description: "This is a sample volume for testing",
+	}
+
+	atc, err := fv.UpdateVolume(volID, &vol)
```
leonwanghui: This ```atc``` means the pointer of attachment resource, so you need to modify it here.

client/volume_test.go
```diff
@@ -439,3 +470,33 @@ func TestDeleteVolumeSnapshot(t *testing.T) {
...
 }
+
+func TestUpdateVolumeSnapshot(t *testing.T) {
+	var snpID = "bd5b12a8-a101-11e7-941e-d77981b584d8"
+	snp := model.VolumeSnapshotSpec{
+		Name:        "sample-snapshot-01",
+		Description: "This is the first sample snapshot for testing",
+	}
+
+	atc, err := fv.UpdateVolumeSnapshot(snpID, &snp)
```
leonwanghui: also here, suggest you change to ```result```

leonwanghui:
LGTM

PR #196, opened 2017-12-19
[WIP] Add opensds ansible installer

leonwanghui:
In this patch, I created a ```ansible``` package to install opensds and backend services through [ansible ](https://github.com/ansible/ansible) tool. Currently I have finished the test of opensds installation with lvm backend, for ceph backend I only finished the code but not testing work. I'll keep you informed my latest work.


leonwanghui:
I will merge this patch now, and we can fix bugs by submitting other patches.

PR #202, opened 2017-12-20
Update ceph playbooks

leonwanghui:


leonwanghui:
This patch is temporarily merged for testing.

PR #218, opened 2017-12-26
Add cli commands for "update volume" and "update volume snapshot"

PengYiOfOpenSDS:


osdsctl/cli/volume.go
```diff
@@ -164,3 +173,25 @@ func volumeDeleteAction(cmd *cobra.Command, args []string) {
...
+	}
+
+	vol := &model.VolumeSpec{
+		Name:        volName,
+		Description: volDesp,
+	}
+
+	resp, err := client.UpdateVolume(args[0], vol)
+	if err != nil {
+		fmt.Println(err)
```
xxwjj: This error infomation should print to stderr.  You can use the fmt.Fprintln(os.Stderr, err) instead of it .

Issue #294, opened 2018-03-08
Add authentication and request filter in osdsctl and client package


leonwanghui:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
According to keystone design (#289) and API request filter support (#284), opensds api-server has supported authentication and request filter features, but as for client and cli tool, these features also need to be supported.

**What you expected to happen**:
Add keystone authentication and api request filter support in cli and client tool.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Hotpot(release/branch) version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

leonwanghui:
@xxwjj @lasshpanzer This issue is releated to PRs you have submitted, please take a look at it.

xxwjj:
@leonwanghui okay,  got it. That's what I'm doing.

leonwanghui:
@lasshpanzer Please take a look at this issue, because there are some issues when I test `osdsctl volume attachment list`.

PR #295, opened 2018-03-09
add multi-tenant feature

xxwjj:
Add multi-tenant feature

Fxied #294 
if you use cli,  some EVNs need to be mentioned:

1. OPENSDS_AUTH_STRATEGY: The authentication strategy, the value can be keystone or noauth, default is  noauth.
2. OPENSDS_TENANT_ID: If you authentication strategy that you specified is onauth, you can use this ENV that set different uuid to test multi-tenant.
3. if the strategy is keystone, the following EVNs need to be setted.The usage is the same as openstack cli.
OS_AUTH_URL
OS_USERNAME
OS_PASSWORD
OS_TENANT_NAME
OS_PROJECT_NAME
OS_USER_DOMAIN_ID

@leonwanghui @xiangrumei @xing-yang  @wisererik 

**Release note**:

```release-note
```


pkg/model/pool.go
```diff
@@ -24,6 +24,10 @@ package model
 // atomic pool and can be abstracted from any storage platform.
 type StoragePoolSpec struct {
 	*BaseModel
+	// The uuid of project
+	// + readOnly
+	ProjectId string `json:"projectId"`
```
leonwanghui: Should be changed to `TenantId`

pkg/model/volume.go
```diff
@@ -69,6 +69,7 @@ type VolumeSpec struct {
 // VolumeAttachmentSpec is a description of volume attached resource.
 type VolumeAttachmentSpec struct {
 	*BaseModel
+
 	// The uuid of the project that the volume belongs to.
```
leonwanghui: `TenantId`

client/receiver.go
```diff
@@ -16,39 +16,70 @@ package client
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
+	"github.com/astaxie/beego/httplib"
```
leonwanghui: Please use `go fmt` to update the code style.

pkg/model/pool.go
xxwjj: I will commit a another PR to solve this issue.

xxwjj:
@leonwanghui  The usage of client, you can find an example in osdslet/cli/cli.go.  You just need to fill up the paramters in KeystonAuthOptions structrue, and assign this option to Config structrue. then you can new a opensds client.

Issue #290, opened 2018-03-08
Scheduler Enhancement


PengYiOfOpenSDS:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
> /kind feature

**What happened**:
1. In the current version, opensds does not have a filtering framework that manages backend storage capabilities.
2. In the current opensds version, users cannot customize the required pool when they create a volume.

**What you expected to happen**:
1. opensds implements a filtering framework for managing backend storage capabilities.
2. In the future, users can customize the required pool when creating a volume.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
Implement the required functionality by modifying filter.go and selector.go.

**Environment**:
- Hotpot(release/branch) version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #297, opened 2018-03-09
scheduler echancement

PengYiOfOpenSDS:

Solve Issue [#290](https://github.com/opensds/opensds/issues/290)

**Release note**:

```release-note
```


pkg/controller/selector/selector.go
```diff
@@ -59,7 +48,7 @@ func (s *selector) SelectSupportedPool(tags map[string]interface{}) (*model.Stor
 		return nil, err
 	}
 
-	supportedPools, err := filterChain.Handle(tags, pools)
+	supportedPools, err := SelectSupportedPools(1, tags, pools)
```
leonwanghui: Can you explain why you select one pool from `SelectSupportedPools`?

pkg/controller/selector/filter.go
```diff
@@ -15,191 +15,304 @@
...
-type CapacityFilter struct {
-	Next Filter
+	return true, nil
 }
 
-func (filter *CapacityFilter) Handle(request map[string]interface{}, pools []*model.StoragePoolSpec) ([]*model.StoragePoolSpec, error) {
-
-	size, ok := request["size"]
+// match ...
+func match(key string, value interface{}, reqValue interface{}) (bool, error) {
```
leonwanghui: I think we should have a detailed discussion about the filter rule, since it's related to capability report ([#281](https://github.com/opensds/opensds/pull/281)) feature.

pkg/controller/selector/filter.go
```diff
@@ -15,191 +15,304 @@
...
+	"strings"
 
+	log "github.com/golang/glog"
 	"github.com/opensds/opensds/pkg/model"
 )
 
-type Filter interface {
-	Handle(request map[string]interface{}, pools []*model.StoragePoolSpec) ([]*model.StoragePoolSpec, error)
+// StructToMap ...
+func StructToMap(structObj interface{}) (map[string]interface{}, error) {
```
leonwanghui: IMO, it's a public method which could be used by other packages. So it could be better if you move this method to `utils/utils.go`, or you can create a `utils` package under selector.

pkg/controller/selector/filter.go
```diff
@@ -15,191 +15,304 @@
...
+	return result, nil
 }
 
-type AZFilter struct {
-	Next Filter
+// Epsilon ...
+const Epsilon float64 = 0.00000001
+
+// IsFloatEqual ...
+func IsFloatEqual(a, b float64) bool {
```
leonwanghui: Same to `StructToMap` method.

pkg/controller/controller.go
```diff
@@ -88,7 +89,7 @@ func (c *Controller) CreateVolume(in *model.VolumeSpec) (*model.VolumeSpec, erro
 	} else {
 		filterRequest = make(map[string]interface{})
 	}
-	filterRequest["size"] = in.Size
+	filterRequest["freeCapacity"] = ">= " + strconv.Itoa(int(in.Size))
```
leonwanghui: The filter request is too limited, you can add more filter parameter according to the definition of [PoolProperties](https://github.com/opensds/opensds/blob/development/contrib/drivers/utils/config/config.go).

pkg/controller/selector/filter.go
```diff
@@ -15,191 +15,304 @@
...
+	if isIn {
+		return true, nil
+	}
+
+	return false, nil
 }
 
-func (filter *ThinFilter) Handle(request map[string]interface{}, pools []*model.StoragePoolSpec) ([]*model.StoragePoolSpec, error) {
+// IsEqual ...
+func IsEqual(key string, value interface{}, reqValue interface{}) (bool, error) {
```
leonwanghui: This is also a public method.

pkg/controller/selector/filter.go
xing-yang: I believe this design follows Cinder's scheduler. So it is actually not necessary to specify all the parameters in the capability reporting in the scheduler. Matching should be done automatically between the profile and the capabilities reported by the backend. That being said, we do need a design doc that has detailed explanations on how this work.

pkg/controller/selector/selector_test.go
```diff
@@ -101,3 +102,732 @@ var (
...
+	if !ok {
+		t.Errorf("Expected ok, get %v", ok)
+	}
+
+	_, ok = poolMap["extras.thin"]
+	if !ok {
+		t.Errorf("Expected ok, get %v", ok)
+	}
+}
+
```
xing-yang: Please take a look of https://github.com/opensds/opensds/pull/281 and add tests accordingly.

pkg/controller/selector/selector.go
PengYiOfOpenSDS: Consistent with the previous logic, the specific reason can refer to the previous comment of the current function "Now, we just return the firt supported pool which will be improved in the future."

pkg/controller/selector/filter.go
PengYiOfOpenSDS: The design document is being written.

pkg/controller/controller.go
PengYiOfOpenSDS: The parameters of filter can be arbitrarily filled.

pkg/controller/selector/selector_test.go
PengYiOfOpenSDS: This is related to StoragePoolSpec in opensds\pkg\model\pool.go and has nothing to do with  #281

PR #309, opened 2018-03-13
replace project id with tenant id

### LINKED ISSUE #308: Replace project id with tenant id.

> /kind bug

**What happened**:
According to our previous discussion,we should use tenant id instead of project id.  So I summit this issue to finnish it.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

xxwjj:
Replace project id with tenant id.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #308 

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/filter/auth/keystone.go
+++ b/pkg/api/filter/auth/keystone.go
@@ -93,7 +93,6 @@ func (k *Keystone) setPolicyContext(r tokens.GetResult) error {
 	}
 
 	ctx := &context.Context{
-		ProjectId:      project.ID,
 		TenantId:       project.ID,
 		Roles:          roleNames,
 		UserId:         user.ID,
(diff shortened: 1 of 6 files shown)


pkg/api/filter/auth/keystone.go
```diff
@@ -93,7 +93,6 @@ func (k *Keystone) setPolicyContext(r tokens.GetResult) error {
 	}
```
leonwanghui: Should we change `ExtractProject` method to `ExtractTenant`?

xxwjj: No, we cann't. The `ExtrackProject` is one method of third library whose name is gohpercloud. There isn't any method named `ExtractTenant`

leonwanghui: Got it.

PR #311, opened 2018-03-14
Capabilities properties design of storage pool

leonwanghui:

design spec [#6](https://github.com/opensds/design-specs/pull/6)

@xing-yang @xxwjj 

**Release note**:

```release-note
```


leonwanghui:
It's odd that just one of ci job was failed to do unit test, but it works when I test it locally.

contrib/drivers/ceph/ceph.go
```diff
@@ -555,17 +555,16 @@ func (d *Driver) getPoolsAttr() (map[string][]string, error) {
...
 
-func (d *Driver) buildPoolParam(line []string, proper PoolProperties) *map[string]interface{} {
-	param := BuildDefaultPoolParam(proper)
-
-	param["redundancyType"] = line[poolType]
-	if param["redundancyType"] == "replicated" {
-		param["replicateSize"] = line[poolTypeSize]
+func (d *Driver) buildPoolExtras(line []string, extras model.StoragePoolExtraSpec) model.StoragePoolExtraSpec {
+	extras.Advanced["redundancyType"] = line[poolType]
+	if extras.Advanced["redundancyType"] == "replicated" {
```
xing-yang: Is "replicated" a Ceph specific redundancy type?  If there are terms like this used in multiple places, it maybe better to use constants to avoid typo.

pkg/controller/selector/filter_test.go.old
```diff
@@ -1,381 +1,381 @@
...
-	for _, testCase := range testCases {
-		result, _ := SelectSupportedPools(len(testCase.pools), testCase.request,
-			testCase.pools)
-
-		if !reflect.DeepEqual(result, testCase.expected) {
-			t.Errorf("Expected %v, get %v", testCase.expected, result)
-		}
-	}
-}
+// Copyright (c) 2017 Huawei Technologies Co., Ltd. All Rights Reserved.
```
xing-yang: Why is this one named filter_test.go.old?

pkg/controller/selector/selector_test.go.old
```diff
@@ -1,815 +1,831 @@
...
-
-	if ExpectedErr != err.Error() {
-		t.Errorf("Expected %v, get %v", ExpectedErr, err)
-	}
-
-	if nil != supportedPools {
-		t.Errorf("Expected nil, get %v", supportedPools[0])
-	}
-}
+// Copyright (c) 2017 Huawei Technologies Co., Ltd. All Rights Reserved.
```
xing-yang: selector_test.go.old?

pkg/model/pool.go
```diff
@@ -53,12 +53,27 @@ type StoragePoolSpec struct {
...
+	// IOConnectivity represents some suggested IO connectivity capabilities.
+	IOConnectivity IOConnectivityLoS `json:"ioConnectivity,omitempty" yaml:"ioConnectivity,omitempty"`
+	// DataProtection represents some suggested data protection capabilities.
+	DataProtection DataProtectionLos `json:"dataProtection,omitempty" yaml:"dataProtection,omitempty"`
+
+	// Besides those basic suggested pool properties above, vendors can configure
+	// some advanced features (diskType, IOPS, throughout, latency, etc)
+	// themselves, all these properties can be exposed to controller scheduler
+	// and filtered by selector in a extensible way.
+	Advanced map[string]interface{} `json:"advanced,omitempty" yaml:"advanced,omitempty"`
```
xing-yang: How about rename "Advanced" to "ExtraProperties" or something else?

contrib/ansible/group_vars/ceph/ceph.yaml
```diff
@@ -1,22 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+configFile: /etc/ceph/ceph.conf
+pool:
+  rbd: # change pool name same to ceph pool, but don't change it if you choose lvm backend
+    storageType: block
```
xing-yang: Is there any validation for the storageType?  What are the supported storageType?  block, file, object?

pkg/model/pool.go
leonwanghui: We already defined `StoragePoolExtraSpec`, it may result in some confusions if we add another `ExtraProperties`.

contrib/ansible/group_vars/ceph/ceph.yaml
leonwanghui: It's discusssed a long time ago, and I just keep it to be compatible, but for its meaning IMO it's about block pool, file pool and object pool such stuffs. For detailed information, @wisererik Erik can solve your question.

contrib/drivers/ceph/ceph.go
leonwanghui: So in ceph every pool can be assigned with two redundancy types: `replicated` and `raid`, and I will add two constants to define them.

pkg/model/pool.go
xing-yang: "Advanced" seems to say IOPS and BWS are not advanced because they are not under this category.  Anyway we can figure out a better name later.

xing-yang:
lgtm

PR #326, opened 2018-03-23
volume, snapshot, attachment statemachine and synchronous to asynchronous

### LINKED ISSUE #287: State machines for volume, snapshot, attachment, dock(agent), pool
> /kind feature
Design proposal is submitted here: https://github.com/opensds/design-specs/pull/4

**What happened**:
There is no mechanism to show the real-time status and state transitions of volume, snapshot, attachment, dock and pool in the current opensds version.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lasshpanzer:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #287 

**Release note**:

```release-note
```


contrib/drivers/ceph/ceph.go
```diff
@@ -234,40 +265,28 @@ func (d *Driver) getSize(img *rbd.Image) int64 {
...
-		Size: d.getSize(img),
-	}, nil
+		return &model.VolumeSpec{
+			BaseModel: &model.BaseModel{
+				Id: name.GetUUID(),
+			},
+			Name: name.GetName(),
+			Size: d.getSize(img),
+		}, nil
+	*/
```
xing-yang: Why do we need to keep this block of commented-out code?

contrib/drivers/ceph/ceph.go
```diff
@@ -382,25 +408,25 @@ func (d *Driver) visitSnapshot(snapID string, fn func(imgName string, img *rbd.I
...
-			if snapName == snapInfo.Name {
-				return fn(imgName, img, &snapInfo)
+			name := ParseName(snapInfo.Name)
+			if snapID == name.GetUUID() {
+				return fn(in, img, &snapInfo)
 			}
 		}
 	}
-	return fmt.Errorf("Not found the snapshot(%s)", snapID)
+	reason := fmt.Sprintf("Not found the snapshot(%s)", snapID)
```
xing-yang: s/Not found the snapshot(%s)/Snapshot (%s) not found.

contrib/drivers/ceph/ceph.go
```diff
@@ -382,25 +408,25 @@ func (d *Driver) visitSnapshot(snapID string, fn func(imgName string, img *rbd.I
...
+	return errors.New(reason)
 }
 
 func (d *Driver) PullSnapshot(snapID string) (*model.VolumeSnapshotSpec, error) {
 	return nil, fmt.Errorf("Ceph PullSnapshot has not implemented yet.")
 	/*
-		if err := d.init(); err != nil {
+		if err := d.initConn(); err != nil {
```
xing-yang: Again why is this block of code commented out and the error message stated that this method is not implemented?

pkg/api/volume.go
```diff
@@ -52,8 +53,7 @@ func (this *VolumePortal) CreateVolume() {
 		return
 	}
 
-	// Call global controller variable to handle create volume request.
-	result, err := controller.Brain.CreateVolume(c.GetContext(this.Ctx), &volume)
+	result, pol, profile, err := controller.Brain.AsynCreateVolume(c.GetContext(this.Ctx), &volume)
```
xing-yang: What is "pol"?  Suggest renaming it to be more readable.

pkg/api/volume.go
```diff
@@ -484,14 +533,26 @@ func (this *VolumeSnapshotPortal) CreateVolumeSnapshot() {
...
-	result, err := controller.Brain.CreateVolumeSnapshot(c.GetContext(this.Ctx), &snapshot)
+	result, err := controller.Brain.AsynCreateVolumeSnapshot(c.GetContext(this.Ctx), &snapshot)
 	if err != nil {
 		reason := fmt.Sprintf("Create volume snapshot failed: %s", err.Error())
 		this.Ctx.Output.SetStatus(model.ErrorBadRequest)
 		this.Ctx.Output.Body(model.ErrorBadRequestStatus(reason))
 		log.Error(reason)
 		return
 	}
+	log.Info("snap的信息是", &snapshot)
```
xing-yang: Please don't write Chinese in the code.

pkg/api/volume.go
```diff
@@ -484,14 +533,26 @@ func (this *VolumeSnapshotPortal) CreateVolumeSnapshot() {
...
+	log.Info("snap的结果是", result)
```
xing-yang: Same here.

pkg/api/volume.go
```diff
@@ -74,6 +74,18 @@ func (this *VolumePortal) CreateVolume() {
...
+	go controller.Brain.CreateVolume(c.GetContext(this.Ctx), result, pol, profile, errchan)
+	if err := <-errchan; err != nil {
+		reason := fmt.Sprintf("Marshal volume created result failed: %s", err.Error())
+		this.Ctx.Output.SetStatus(model.ErrorBadRequest)
+		this.Ctx.Output.Body(model.ErrorBadRequestStatus(reason))
+		log.Error(reason)
+		return
+	}
+
 	return
```
xing-yang: So this CreateVolume function doesn't return after AsynCreateVolume call?  How is this really async?

Does controller.Brain.CreateVolume wait until a volume is really created?

This comment applies to all other "async" functions.

contrib/drivers/ceph/ceph.go
xxwjj: @xing-yang This commented out interface doesn't  used in current version. and this interface will be mofidied in the future,  so we just commented out it here.

xing-yang: It looks like @xxwjj replied here (I got an email notification) but can't find it here:).  If it is not used now but will be used in the future, can you please add a "NOTE" here to explain?  Usually unused code should be removed.  If you want to keep it, please add a NOTE explaining why.

pkg/api/volume.go
```diff
@@ -52,8 +52,10 @@ func (this *VolumePortal) CreateVolume() {
 		return
 	}
 
-	// Call global controller variable to handle create volume request.
-	result, err := controller.Brain.CreateVolume(c.GetContext(this.Ctx), &volume)
+	// NOTE:It will create a volume entry into the database and initialize its status
+	// as "creating". It will not wait for the real volume creation to complete
+	// and will return result immediately.
+	result, pool, profile, err := controller.Brain.CreateVolumeDBEntry(c.GetContext(this.Ctx), &volume)
```
leonwanghui: Like I said, the method between api package and controller package must be unique for every operation, you can't call two methods at the same time, and here are some reasons:
* Api package can't decide whether the operation is sync or not, it's all decided by controller package.
* Later if we decouple api and controller package, then we have to reduce the calling times because in your design it would take two calls if users want to create a volume.

pkg/api/db.go
```diff
@@ -0,0 +1,230 @@
...
+	if in.CreatedAt == "" {
+		in.CreatedAt = time.Now().Format(constants.TimeFormat)
+	}
+	if len(in.AdditionalProperties) == 0 {
+		in.AdditionalProperties = map[string]interface{}{"attachment": "attachment"}
+	}
+	if len(in.ConnectionData) == 0 {
+		in.ConnectionData = map[string]interface{}{"attachment": "attachment"}
+	}
+	if in.Platform == "" {
```
leonwanghui: Why do you add these assertion? From my point of view, it has no meaning to assert these empty fields.

pkg/api/db.go
```diff
@@ -0,0 +1,230 @@
...
+	result, err := db.C.CreateVolumeSnapshot(ctx, snap)
+	if err != nil {
+		log.Error("Error occurred in dock module when create volume snapshot in db:", err)
+		return nil, err
+	}
+	return result, nil
+}
+
+func DeleteVolumeSnapshotDBEntry(ctx *c.Context, in *model.VolumeSnapshotSpec) error {
+	if in.Status != model.VOLUMESNAP_AVAILABLE {
```
leonwanghui: The logic is not right, you need to find the real volume snapshot in database by calling `GetVolumeSnapshot` method first. You can't never find the status of volume snapshot from `in`.

pkg/api/db.go
```diff
@@ -0,0 +1,230 @@
...
+	in.Status = model.VOLUMESNAP_DELETING
+	_, err := db.C.UpdateVolumeSnapshot(ctx, in.Id, in)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+//Just modify the state of the volume to be deleted in the DB, the real deletion in another thread
+func DeleteVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) error {
```
leonwanghui: Same here.

pkg/api/db.go
lasshpanzer: Please see the code yourself.

pkg/api/db.go
lasshpanzer: There is no need to implement async process of Update method. It updates directly at the API level without going though the backend.

PR #328, opened 2018-03-24
add replication dock framework

xxwjj:
Continue with @xing-yang's works and fix some building error.

**Release note**:

```release-note
```


contrib/drivers/drivers.go
```diff
@@ -58,8 +58,6 @@ type VolumeDriver interface {
 	DeleteSnapshot(opt *pb.DeleteVolumeSnapshotOpts) error
 
 	ListPools() ([]*model.StoragePoolSpec, error)
-
-	CreateReplication(opt *pb.CreateReplicationOpts) (*model.ReplicationSpec, error)
```
xing-yang: This is needed for array-based replication, but you can add that later.

pkg/model/replication.go
```diff
@@ -52,4 +47,30 @@ type ReplicationSpec struct {
...
+	ReplicationModel string `json:"replicationModel,omitempty"`
+	// 0 means sync replication.
+	ReplicationPeriod    int    `json:"replicationPeriod,omitempty"`
+	// replication period
+	ReplicationBandwidth int    `json:"replicationBandwidth,omitempty"`
+	// profile id
+	ProfileId            string `json:"profileId,omitempty"`
+}
+
+type FailoverReplicationSpec struct {
```
xing-yang: Why does FailoverReplicationSpec need to be a separate spec from ReplicationSpec?

contrib/drivers/drbd/replication.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2018 Huawei Technologies Co., Ltd. All Rights Reserved.
```
xing-yang: Remove this line from this file.

xing-yang:
I"ll merge this now.  We can talk about FailoverReplicationSpec later.  Thanks.

xing-yang:
lgtm

pkg/model/replication.go
xing-yang: Ok. I think this is fine.  FailoverReplicationSpec is from the request body of the failover_replication API.

PR #332, opened 2018-03-28
Fixed some multi-tenant bugs

xxwjj:
1. Fixed some multi-tenant bugs
2. In noauth case, only the default id is treated as admin role.

**Release note**:

```release-note
```

### DIFF
--- a/pkg/db/drivers/etcd/etcd.go
+++ b/pkg/db/drivers/etcd/etcd.go
@@ -648,7 +648,7 @@ func (c *Client) CreateProfile(ctx *c.Context, prf *model.ProfileSpec) (*model.P
 	if prf.CreatedAt == "" {
 		prf.CreatedAt = time.Now().Format(constants.TimeFormat)
 	}
-
+	prf.TenantId = ctx.TenantId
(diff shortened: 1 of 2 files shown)


pkg/db/drivers/etcd/etcd.go
```diff
@@ -1717,6 +1762,14 @@ func (c *Client) DeleteVolumeSnapshot(ctx *c.Context, snpID string) error {
 	dbReq := &Request{
 		Url: urls.GenerateSnapshotURL(urls.Etcd, ctx.TenantId, snpID),
 	}
+	if IsAdminContext(ctx) {
```
leonwanghui: What do you think if changing into this:
```go
	if IsAdminContext(ctx) {
		snap, err := c.GetVolumeSnapshot(ctx, snpID)
		if err != nil {
			log.Error(err)
			return err
		}
		ctx.TenantId = snap.TenantId
	}
	dbReq := &Request{
		Url: urls.GenerateSnapshotURL(urls.Etcd, ctx.TenantId, snpID),
	}
```

PR #359, opened 2018-04-18
Refatored the LVM ISCSI feature.

xxwjj:
Using persistent configure file instead of command `tgtadm`  to initialize the iscsi  connection.

**Release note**:

```release-note
```

### DIFF
--- a/contrib/drivers/lvm/targets/iscsi.go
+++ b/contrib/drivers/lvm/targets/iscsi.go
@@ -31,7 +31,7 @@ const (
 )
 
 type ISCSITarget interface {
-	CreateISCSITarget(volId, tgtIqn, path, initiator string, chapAuth []string) error
+	CreateISCSITarget(volId, tgtIqn, path, hostIp, initiator string, chapAuth []string) error
 	GetISCSITarget(iqn string) int
 	RemoveISCSITarget(volId, iqn string) error
 	GetLun(path string) int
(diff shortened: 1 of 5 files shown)


contrib/drivers/lvm/targets/iscsi.go
```diff
@@ -21,48 +21,48 @@ import (
...
 	log "github.com/golang/glog"
+	"github.com/opensds/opensds/pkg/utils"
+	"os"
 )
 
 type ISCSITarget interface {
-	CreateISCSITarget() error
-	GetISCSITarget() int
-	RemoveISCSITarget() error
+	CreateISCSITarget(volId, tgtIqn, path, initiator string, chapAuth []string) error
```
leonwanghui: I think it could be better if these parameters can be included in `tgtTarget` struct.

contrib/drivers/lvm/targets/iscsi.go
```diff
@@ -161,91 +156,29 @@ func (t *tgtTarget) GetISCSITarget() int {
...
 
+	os.Remove(tgtConfPath)
 	return nil
 }
 
-func (*tgtTarget) execCmd(cmd []string) (string, error) {
-	ret, err := exec.Command("tgtadm", cmd...).Output()
+func (*tgtTarget) execCmd(name string, cmd ...string) (string, error) {
+	ret, err := exec.Command(name, cmd...).Output()
+	log.V(8).Infoln("Command: ", "tgtadm ", strings.Join(cmd, " "))
```
leonwanghui: Why is `tgtadm` contained in log?

contrib/drivers/lvm/targets/iscsi.go
```diff
@@ -16,69 +16,43 @@ package targets
...
-		log.Error("Fail to exec 'tgtadm' to add lun into iscsi target:", err)
-		return err
-	}
-
-	return nil
+	BindIp     string
+	TgtConfDir string
 }
 
 func (t *tgtTarget) GetLun(path string) int {
```
leonwanghui: Is there special use case for keeping `GetLun` method?

xxwjj: The ConnectionData  need this id.

contrib/drivers/lvm/targets/iscsi.go
xxwjj: Thanks for reminding me.

contrib/drivers/lvm/targets/iscsi.go
leonwanghui: Got it

PR #386, opened 2018-05-15
Modify extend volume api

BaiHuoYu:

design spec [#11](https://github.com/opensds/design-specs/pull/11/files)

**Release note**:

```release-note
```

### DIFF
--- a/client/volume.go
+++ b/client/volume.go
@@ -132,7 +132,7 @@ func (v *VolumeMgr) ExtendVolume(volID string, body ExtendVolumeBuilder) (*model
 	var res model.VolumeSpec
 	url := strings.Join([]string{
 		v.Endpoint,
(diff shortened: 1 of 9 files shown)


client/volume.go
```diff
@@ -132,7 +132,7 @@ func (v *VolumeMgr) ExtendVolume(volID string, body ExtendVolumeBuilder) (*model
 	var res model.VolumeSpec
 	url := strings.Join([]string{
 		v.Endpoint,
-		urls.GenerateNewVolumeURL(urls.Client, v.TenantId, volID+"/action")}, "/")
+		urls.GenerateNewVolumeURL(urls.Client, v.TenantId, volID+"/resize")}, "/")
```
leonwanghui: I would prefer `urls.GenerateNewVolumeURL(urls.Client, v.TenantId, volID, "resize")}, "/")`, and BTW it could be better if define resize as a constants. For example, just define `ResizeAction` as "resize".

PR #387, opened 2018-05-15
Change replication action url from unify name to actual action name

xxwjj:

Change replication action url from unify name to actual action name

**Release note**:

```release-note
```

### DIFF
--- a/client/replication.go
+++ b/client/replication.go
@@ -108,8 +108,7 @@ func (v *ReplicationMgr) EnableReplication(replicaId string) error {
 	url := strings.Join([]string{
 		v.Endpoint,
 		urls.GenerateReplicationURL(urls.Client, v.TenantId, replicaId, "enable")}, "/")
-	body := map[string]interface{}{}
-	return v.Recv(url, "POST", body, nil)
+	return v.Recv(url, "POST", nil, nil)
 }
 
 // EnableReplication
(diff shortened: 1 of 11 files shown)


openapi-spec/swagger.yaml
```diff
@@ -1124,11 +1124,6 @@ paths:
           in: body
```
leonwanghui: Honestly I don't like the empty body defined here, I would prefer to remove `body` object if nothing needs to be passed in the request body.

openapi-spec/swagger.yaml
```diff
@@ -1093,11 +1098,6 @@ paths:
...
```
leonwanghui: Same with what's in disable method.

client/replication.go
```diff
@@ -107,24 +107,24 @@ func (v *ReplicationMgr) UpdateReplication(replicaId string, body ReplicationBui
 func (v *ReplicationMgr) EnableReplication(replicaId string) error {
 	url := strings.Join([]string{
 		v.Endpoint,
-		urls.GenerateReplicationURL(urls.Client, v.TenantId, replicaId, "action")}, "/")
-	body := map[string]interface{}{"enableReplication": map[string]interface{}{}}
+		urls.GenerateReplicationURL(urls.Client, v.TenantId, replicaId, "enable")}, "/")
+	body := map[string]interface{}{}
```
leonwanghui: There is no need to pass empty body here, `nil` is just ok.

client/replication.go
```diff
@@ -107,24 +107,24 @@ func (v *ReplicationMgr) UpdateReplication(replicaId string, body ReplicationBui
...
 }
 
 // EnableReplication
 func (v *ReplicationMgr) DisableReplication(replicaId string) error {
 	url := strings.Join([]string{
 		v.Endpoint,
-		urls.GenerateReplicationURL(urls.Client, v.TenantId, replicaId, "action")}, "/")
-	body := map[string]interface{}{"disableReplication": map[string]interface{}{}}
+		urls.GenerateReplicationURL(urls.Client, v.TenantId, replicaId, "disable")}, "/")
+	body := map[string]interface{}{}
```
leonwanghui: Same with last comment

PR #404, opened 2018-05-24
Fixed a bug  about capacity unit

xxwjj:
1.  Using cmd vgs instead of vgdisplay.
2. Fixed a bug about capacity unit.
3. Using VG UUID to generate pool UUID.

**Release note**:

```release-note
```


contrib/drivers/lvm/lvm_test.go
```diff
@@ -356,4 +358,8 @@ var (
 		- currently set to     256
 		Block device           253:3
 	`
+	sampleVGS = `
```
leonwanghui: Please change the code style here.

xxwjj: We should keep the actual commond out text format.

PR #410, opened 2018-05-26
Add Integrate CI Test--Flow_test.go&&scanvolume.ssh

lijuncloud:
This PR Add Two Parts:
  1、Flow_test.go
        The file is for Integration Test about profile(create/list/detail/delete),volume(create/list/detail/extend/update/delete),attachment(create/list/detail/delete) and snapshot(create/list/detail/update/delete)
   2、scanvolume.sh
      This file is create for CI Test About Create Attachment Test  & Delete Attachment,After Create and Delete a Attachment,The file will login  the target note to scan the volume and check if the volume is really attached/deattached


lijuncloud:
Please Help to Chek the code,Tks

PR #421, opened 2018-05-29
Fxied some bugs in drcontroller

xxwjj:
1.  Clean up all replicationDriverData in volume  after deleting replication.
2.  Modified replication south invoking stratege, the drcontroller will invoke two both sides of nodes.

**Release note**:

```release-note
```


pkg/dock/proto/dock.proto
```diff
@@ -366,6 +368,8 @@ message EnableReplicationOpts {
     string context = 15;
     // The replication metadata
     map<string, string> metadata = 16;
+    // Whether is primary replication
+    bool  isPrimary = 17;
```
leonwanghui: I‘m a bit confused if it's ok to add another field here, is it better to pass ti inside `metadata`?

PR #425, opened 2018-05-29
replication,drbd: initial version

rck:
This implements replication via DRBD9[1]. First you need to create two
volumes (most likely LVM) via "osdsctl volume create" on two different
hosts. After that, a DRBD replication can be created between these two
hosts/volumes.

Currently this is only ready for testing, but in no ways for anything
near production. There are two fundamental limits:

- Shared state in OpenSDS:
  DRBD uses shared information that has to be consistent and available
  on both hosts. This mainly includes the chosen TCP/IP ports and the
  DRBD minor number. As there are no real means to share that
  information in the OpenSDS cluster, this works as follows:
  - use the same port/minor minimum version (drbd.yaml) on both hosts
  - store meta data on CreateReplication() that stores used ports/minors.
  - check which ones are locally used, and chose the next one.
  This can fail in various scenarios, and if that machinery gets out of
  "sync" once, the different hosts will chose differnt ports/minors.
  This is ways to fragile for production use.

  This shared information is attached to the local volumes, which has
  another side effect: If you create a replication for LVM1+LVM2, delete
  that replication, create it again for these two devices, the stored
  meta data still exists, because it got attached to the devices. In
  that case the next valid port/minor number combination is chosen. It
  would be a lot more robust, if we could fail at the point where we
  detect that a volume already has a port/minor combination.
  Unfortunately this is not possible, as it would break the
  create/delete/create again case.

- Export of underlying backing devices:
  Currently, OpenSDS exports created devices for access. For example the
  DRBD backing devices (LVM volumes) are iSCSI exported. Accessing the
  underlying backing devices directly breaks DRBD. This would also
  apply to every layered block devices in Linux (e.g., accessing the
  underlying devices of a DM-crypt device also breaks the encryption
  device; or you also do not access the underlying devices of a software
  RAID. That is simply how the Linux block layer works). OpenSDS is
  currently not able to handle how Linux works, namely layered block
  devices. IMO, this design issue will need to be solved anyways,
  otherwise layering of Linux block devices will not be possible.
  Hopefully fixing the OpenSDS design then fixes this problem for DRBD.

[1] www.linbit.com

**Release note**:

```release-note
```


leonwanghui:
Good job!

xing-yang:
Thanks Roland!

xing-yang:
I opened #426 and #427 to keep track of the two issues mentioned Roland.

xing-yang:
@rck #427 (DRBD: Shared state in OpenSDS) is already fixed by #421.  Please give a try when you get a chance.

PR #433, opened 2018-06-01
add metadata into the whitelist for replication api

xxwjj:
Add metadta into the whitelist for replication api

In csi replication scenario, some info should be stored in replication metadata when in Create Volume steps, and the info can be fetched in ControllerPublish.

**Release note**:

```release-note
```


xing-yang:
Please explain why you need to add metadata in the PR description section.  Thanks.

xxwjj:
@xing-yang  In csi replication scenario ,I need to store some info in replication metadata when in Create Volume steps,then I will get the info in ControllerPublish.

PR #438, opened 2018-06-05
Create volume from snapshot

lixiaoy1:

**BP Create_Volume_From_Snapshot.

This is to implement volume creation from snapshots.

**Release note**:

```release-note
Users can create volumes from snapshots.
```

### DIFF
--- a/contrib/drivers/lvm/lvm.go
+++ b/contrib/drivers/lvm/lvm.go
@@ -38,6 +38,8 @@ const (
 	defaultConfPath   = "/etc/opensds/driver/lvm.yaml"
 	volumePrefix      = "volume-"
 	snapshotPrefix    = "_snapshot-"
+	blocksize         = 4096
+	sizeShiftBit      = 30
 )
 
 type LvInfo struct {
@@ -107,6 +109,24 @@ func (d *Driver) CreateVolume(opt *pb.CreateVolumeOpts) (*model.VolumeSpec, erro
 		}
 		if strings.Contains(line, "LV Status") {
 			lvStatus = strings.Fields(line)[2]
+		}
+	}
+
+	// Copy snapshot to volume
+	var snap = opt.GetSnapshotId()
+	if snap != "" {
+		var snapSize = uint64(opt.GetSnapshotSize())
+		var count = (snapSize<<sizeShiftBit)/blocksize
+		var snapName = snapshotPrefix + snap
+		var snapPath = path.Join("/dev", polName, snapName)
+		if _, err := d.handler("dd", []string{
+			"if=" + snapPath,
+			"of=" + lvPath,
+			"count=" + fmt.Sprint(count),
+			"bs=" + fmt.Sprint(blocksize),
+		}); err != nil {
+			log.Error("Failed to create logic volume:", err)
+			return nil, err
 		}
 	}
 
--- a/pkg/api/db.go
+++ b/pkg/api/db.go
@@ -44,6 +44,23 @@ func CreateVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) (*model.VolumeSpe
 		log.Error(errMsg)
 		return nil, errors.New(errMsg)
 	}
+	if in.SnapshotId != "" {
+		snap, err := db.C.GetVolumeSnapshot(ctx, in.SnapshotId)
+		if err != nil {
+			log.Error("Get snapshot failed in create volume method: ", err)
+			return nil, err
+		}
+		if snap.Status != model.VolumeSnapAvailable {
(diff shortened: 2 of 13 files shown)


lixiaoy1:
@leonwanghui Rebased on development branch.

leonwanghui:
It seems that the unit test is not right, please take a look

contrib/drivers/lvm/lvm.go
```diff
@@ -110,6 +112,24 @@ func (d *Driver) CreateVolume(opt *pb.CreateVolumeOpts) (*model.VolumeSpec, erro
 		}
 	}
 
+        // Copy snapshot to volume
+        var snap = opt.GetSnapshotId()
```
wisererik: the default tab size should be 4.

wisererik:
Hi @lixiaoy1  The API document should be updated which is located in [opensds/openapi-spec/swagger.yaml](https://github.com/opensds/opensds/tree/master/openapi-spec).

pkg/api/db.go
```diff
@@ -44,6 +44,23 @@ func CreateVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) (*model.VolumeSpe
...
 		return nil, errors.New(errMsg)
 	}
+        if in.SnapshotId != "" {
+                snap, err := db.C.GetVolumeSnapshot(ctx, in.SnapshotId)
+                if err != nil {
+                        log.Error("Get snapshot failed in create volume method: ", err)
+                        return nil, err
+                }
+                if snap.Status != model.VolumeSnapAvailable {
+                        var errMsg = "Only the snapshot is available, the volume can be created"
```
xing-yang: s/Only the snapshot/Only if the snapshot

pkg/controller/controller.go
```diff
@@ -120,6 +150,8 @@ func (c *Controller) CreateVolume(ctx *c.Context, in *model.VolumeSpec, errchanV
 		AvailabilityZone: in.AvailabilityZone,
 		ProfileId:        profile.Id,
 		PoolId:           polInfo.Id,
+                SnapshotId:       in.SnapshotId,
+                SnapshotSize:     snapSize,
```
xing-yang: Is it necessary to pass "SnapshotSize" as a parameter to CreateVolumeOpts and compare that to volume size when creating the db entry?   Can you do the comparison here and if size is not right, don't proceed to create a db entry?

pkg/dock/proto/dock.proto
```diff
@@ -112,6 +112,8 @@ message CreateVolumeOpts {
     map<string, string> replicationDriverData = 13;
     // The uuid of replication
     string replicationId = 14;
+    // The size of snapshot
+    int64 snapshotSize = 15;
```
xing-yang: See my previous comment.  It seems not necessary to add this snapshotSize in CreateVolumeOpts.  We can compare snapshot size with volume size earlier and bail out if it is not right.

contrib/drivers/lvm/lvm.go
lixiaoy1: Done.

pkg/api/db.go
lixiaoy1: Done

pkg/controller/controller.go
lixiaoy1: We allow creating a volume from a snapshot whose size is bigger than the snapshot. So it needs the size of the snapshot in the driver, for example, lvm.go. Or else it needs to query the size of the snapshot in lvm.go. How do you think?

contrib/drivers/lvm/lvm.go
```diff
@@ -110,6 +112,24 @@ func (d *Driver) CreateVolume(opt *pb.CreateVolumeOpts) (*model.VolumeSpec, erro
 		}
 	}
 
+	// Copy snapshot to volume
+	var snap = opt.GetSnapshotId()
```
leonwanghui: I think the operation of create volume and create volume from snapshot can not be executed at the same time, so you should add a switch flag to identify which operation should be performed.

Maybe I didn't understand, please explain what is `dd` command used for.

lixiaoy1: They will both be executed when creating from snapshots. lvcreate at first creates an lv, and then dd copies data from source snapshot lv to the new lv.  LVM has no command to create a thick lv from a snapshot directly.

leonwanghui: Got it, thanks!

pkg/controller/controller.go
xing-yang: Got it.  Thanks!

PR #445, opened 2018-06-06
resolve volume can not be deleted error. Store all volume info into d…

lasshpanzer:
…b before rpc call driver to create volume.

**Release note**:

```release-note
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strconv"
 
 	log "github.com/golang/glog"
 	c "github.com/opensds/opensds/pkg/context"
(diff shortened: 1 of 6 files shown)


pkg/controller/controller.go
```diff
@@ -62,97 +61,19 @@ type Controller struct {
...
-	if err != nil {
-		log.Error("When search supported dock resource:", err.Error())
-		if errUpdate := db.C.UpdateStatus(ctx, in, model.VolumeError); errUpdate != nil {
-			errchanVolume <- errUpdate
-			return
-		}
-		errchanVolume <- err
-		return
-	}
+func (c *Controller) CreateVolume(ctx *c.Context, in *model.VolumeSpec, poolName string, dockInfo *model.DockSpec, snapSize int64) error {
```
leonwanghui: I don't get the point why do you move the logic out of controller.

PR #461, opened 2018-06-12
Change the default port of dashboard service

wisererik:
 Because keystone uses 80 as default port in apach2 service and hyperkube process of kubernetes also uses 8080 port, so changing the default port to 8088 of dashboard service.
**Which issue this PR fixes**:

**Release note**:

```release-note
```


xing-yang:
Looks like you need a rebase before this can be merged.

PR #462, opened 2018-06-12
Fixed keystone authentication failed issue

xxwjj:
1. Fixed keystone authentication failed issue

**Release note**:

```release-note
```


pkg/api/filter/auth/keystone.go
```diff
@@ -84,27 +83,28 @@ func (k *Keystone) setPolicyContext(r tokens.GetResult) error {
...
-	ctx := &context.Context{
-		TenantId:       project.ID,
-		Roles:          roleNames,
-		UserId:         user.ID,
-		IsAdminProject: project.Name == "admin",
+	param := map[string]interface{}{
+		"TenantId":       project.ID,
+		"Roles":          roleNames,
+		"UserId":         user.ID,
+		"IsAdminProject": project.Name == "admin",
```
xing-yang: This check should be case in-sensitive.  "admin" could be "Admin", etc.

leonwanghui: @xing-yang You mean `admin` is equal to `Admin`?

pkg/context/context.go
```diff
@@ -19,10 +19,9 @@ package context
 
 import (
 	"encoding/json"
-	"reflect"
-
 	"github.com/astaxie/beego/context"
 	"github.com/golang/glog"
+	"reflect"
```
leonwanghui: Please use `go fmt` to update it.

pkg/api/filter/auth/keystone.go
xing-yang: I meant it is possible the admin user is named "Admin" instead of "admin".  Both should be fine.  By the way, is there a better way of checking this rather than a string comparison?  Is there a method we can call?

xxwjj: using strings.Tolower may be better.

PR #484, opened 2018-06-29
Switch to using dep for dependency management

stmcginnis:
Glide is no longer actively developed and the Go community
recommendation is to move to the "dep" project for dependency
management.

This migrates our existing glide usage to dep and picks up
the latest vendor package updates.


leonwanghui:
Awesome, I have considered it long time ago : )

Issue #495, opened 2018-07-04
If the volume being extended has a snapshot and the driver is lvm, deactivate the snapshot first.


BaiHuoYu:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
> /kind bug

**What happened**:
If the volume being extended has a snapshot and the driver is lvm, it has not been extended successfully.

**What you expected to happen**:
If the volume being extended has a snapshot and the driver is lvm, deactivate the snapshot first.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
None

**Environment**:
- Hotpot(release/branch) version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #496, opened 2018-07-04
If the volume being extended has a snapshot and the driver is lvm, deactivate the snapshot first.

BaiHuoYu:

Solve Issue [#495](https://github.com/opensds/opensds/issues/495)

**Release note**:

```release-note
```


contrib/drivers/lvm/lvm.go
```diff
@@ -202,13 +207,65 @@ func (d *Driver) lvHasSnapshot(lvPath string) bool {
...
 	return info[0] == 'o' || info[0] == 'O'
 }
 
+func (d *Driver) getActiveSnapshotsPathsOfLv(lvPath string) ([]string, error) {
+	array := strings.Split(lvPath, "/")
+	lvName := array[len(array)-1]
+	var snapshotsPaths []string
+
+	args := []string{}
+	info, err := d.handler("lvdisplay", args)
```
stmcginnis: It would be much more concise to use use "lvdisplay --noheading -C -o Attr" instead of the full default human readable output of lvdisplay. That would give

BaiHuoYu: If I use "lvdisplay --noheading -C -o Attr", I can't get the value of "LV snapshot status".

wisererik:
please add integration test to verify the issue: 
1. create volume
2. create snapshot
3. extend volume size

contrib/drivers/lvm/lvm.go
stmcginnis: OK, thanks for pointing that out. It seems like there should be a better way to match up snapshot volumes with their parents, but I don't know off the top of my head.

One issue with this though - I don't believe it will work with thin provisioned volumes. It looks like in that case it is not "LV snapshot status" but rather "LV Thin origin name  <name>"

BaiHuoYu: Because my ultimate goal is to deactivate the snapshot with the status "active", so I use "LV snapshot status" instead of "LV Thin origin name".

leonwanghui:
Hi @BaiHuoYu , can you give some clues on the status of this PR?

PR #502, opened 2018-07-19
Standardize some constants defined for all storage drivers

leonwanghui:
In southbound drivers module, it's too disorganized that we define many string constants randomly. So this patch is designed for keeping all constants defined in drivers module at one place.

**Release note**:

```release-note
```

### DIFF
--- a/contrib/connector/fc/fc.go
+++ b/contrib/connector/fc/fc.go
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	fcDriver = "fc"
+	fcDriver = "fibre_channel"
 )
 
 type FC struct {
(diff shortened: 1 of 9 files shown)


leonwanghui:
@xing-yang @stmcginnis @wisererik Please take a look at this PR, thanks!

contrib/connector/fc/fc.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	fcDriver = "fc"
+	fcDriver = "fibre_channel"
```
stmcginnis: Could probably just import from constants.go, but this looks fine.

leonwanghui:
I planned to do that, but it’s a bit odd if the code in connector package
imports from constants in drivers module.

Just my thought.

Sean McGinnis <notifications@github.com>于2018年7月20日 周五19:59写道：

stmcginnis:
True. I wonder if we should move those constants up to a higher, more
general level.

Looks just fine as is, but maybe something to think about if we end up
eventually having more constants that are common across different areas of
the code.

Issue #432, opened 2018-06-01
Profile to volume_type mapping


xing-yang:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
Currently there is no mapping between OpenSDS profile and Cinder volume_type.

**What you expected to happen**:
1. When creating a new profile in OpenSDS, OpenSDS should create a corresponding volume_type in Cinder if Cinder is a backend.
2. When using an existing Cinder setup, OpenSDS should create a profile that is mapped to an existing volume_type in Cinder.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Hotpot(release/branch) version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

leonwanghui:
Assigned to @lijuncloud

xing-yang:
The default Cinder volume type is always used.  We should pass in a cinder volume type at volume creation time.  That volume type name/uuid should be in the profile.

In the following code, volume type should be passed in, in addition to name, description, and size:
https://github.com/opensds/opensds/blob/master/contrib/drivers/openstack/cinder/cinder.go#L172

When creating a profile that will be used for cinder, cinder volume type name/uuid should be part of the custom properties.

skdwriting:
@leonwanghui did we consider this? I am moving to backlog. Please comment, if any change needed.

PR #518, opened 2018-08-18
Get All StoragePools FreeCaps by VolumeType

lijuncloud:
  
Modification point：
  1、contrib/drivers/openstack/cinder/cinder.go
 2、vendor/github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/schedulerstats/results.go
  3、vendor/github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/schedulerstats/url.go
  4、pkg/model/volumetype.go


pkg/model/volumetype.go
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2017 Huawei Technologies Co., Ltd. All Rights Reserved.
```
xxwjj: The voume type is only used in cinder driver, it's not very good to put volumetype.go file in pkg/model directory. Puting it into cinder.go file maybe better.

leonwanghui:
#432

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -22,6 +22,9 @@ Go SDK.
 package cinder
 
 import (
+	"io/ioutil"
```
leonwanghui: Please use `go fmt' to improve the code style.

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -76,6 +81,36 @@ type CinderConfig struct {
...
+	Pools []StoragePool `json:"pools"`
+}
+
+type StoragePool struct {
+	Name         string       `json:"name"`
+	Capabilities Capabilities `json:"capabilities"`
+}
+
+type Capabilities struct {
+	FreeCapacityGB  float64 `json:"free_capacity_gb"`
```
leonwanghui: Can you explain why the type is `float64`? Because it could cause some type changes during the conversion.

lijuncloud: I copy the struct from "vendor/github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/schedulerstats/results.go",it was defined in this file.but I can't use it directly as some type of the struct is different from the result of the Interface call。

contrib/drivers/openstack/cinder/cinder.go
lijuncloud: I have formatted the file，Thank you for your opinions

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -22,6 +22,9 @@ Go SDK.
 package cinder
 
 import (
+	"io/ioutil"
+
```
leonwanghui: This blank line should be removed

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -37,6 +40,8 @@ import (
 	"github.com/opensds/opensds/pkg/model"
 	"github.com/opensds/opensds/pkg/utils/config"
 	"github.com/satori/go.uuid"
+
```
leonwanghui: This blank line should be removed

contrib/drivers/openstack/cinder/cinder.go
lijuncloud: Have remove it

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -76,6 +79,36 @@ type CinderConfig struct {
 	Pool        map[string]PoolProperties `yaml:"pool,flow"`
 }
 
+//ListPoolOpts
```
xing-yang: nit: Add a space after //

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -76,6 +79,36 @@ type CinderConfig struct {
...
+type ListPoolOpts struct {
+	// ID of the tenant to look up storage pools for.
+	TenantID string `q:"tenant_id"`
+	// Whether to list extended details.
+	Detail bool `q:"detail"`
+	//Volume_Type of the StoragePool
```
xing-yang: nit: Add a space after //

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -76,6 +79,36 @@ type CinderConfig struct {
...
+	Volume_Type string `q:"volume_type"`
```
xing-yang: It is possible to map multiple volume types to the same pool.  If you only have one volume type here, how do you capture other volume types also mapped to this pool if there are any?

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -76,6 +79,36 @@ type CinderConfig struct {
...
+}
+
+//Struct of Pools listed by volumeType
```
xing-yang: space after //

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -381,40 +414,52 @@ func (d *Driver) DeleteSnapshot(req *pb.DeleteVolumeSnapshotOpts) error {
...
+			return nil, err
+		}
+		var StoragePools PoolArray
+		json.Unmarshal(polpage, &StoragePools)
+		var freeCaps int64 = 0
+		var totalCaps int64 = 0
+		var pol *model.StoragePoolSpec
+		for _, page := range StoragePools.Pools {
+			freeCaps = freeCaps + int64(page.Capabilities.FreeCapacityGB)
+			totalCaps = totalCaps + int64(page.Capabilities.TotalCapacityGB)
```
xing-yang: It is int64 here but it is FreeCapacityGB  float64 `json:"free_capacity_gb"` in cinder/cinder.go?

contrib/drivers/openstack/cinder/cinder.go
lijuncloud: I get all volume types from the customer's configuration file:"/etc/opensds/driver/cinder.yaml"，Then I add them to a array  and Traversing this array, At last,I get pools by each volume type and add pools to another array and return all pools.

contrib/drivers/openstack/cinder/cinder.go
xing-yang: Can you fix this one too?

PR #521, opened 2018-08-27
Fixed some dorado driver bugs

xxwjj:
1. Fixed  dorado replication ReplicationPeriod UNITE issue.
2. Fixed dorado create volume always in pool 0 issue.

**Release note**:

```release-note
```

### DIFF
--- a/contrib/drivers/huawei/dorado/replication.go
+++ b/contrib/drivers/huawei/dorado/replication.go
@@ -64,7 +64,7 @@ func (r *ReplicationDriver) CreateReplication(opt *pb.CreateReplicationOpts) (*m
 	}
 	pLunId := opt.PrimaryReplicationDriverData[KLunId]
 	sLunId := opt.SecondaryReplicationDriverData[KLunId]
-	replicationPeriod := strconv.FormatInt(opt.ReplicationPeriod, 10)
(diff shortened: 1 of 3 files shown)


contrib/drivers/huawei/dorado/replication.go
```diff
@@ -269,7 +269,7 @@ func (r *ReplicaPairMgr) CreateRemoteLun(localLun *Lun) (*Lun, error) {
 	if err != nil {
 		return nil, err
 	}
-	rmtLun, err := r.remoteClient.CreateVolume(localLun.Name, sector, localLun.Description)
+	rmtLun, err := r.remoteClient.CreateVolume(localLun.Name, sector, localLun.Description, "0")
```
wisererik: Will be remote lun be created on pool 0?  What about if pool 0 doesn't exist?

xxwjj: These code is not used and  will be removed next version, modified it here is just for building error.

xxwjj: This code is design for creating remote lun when user not provide the remote lun id. But in current version, user should create local lun and remote lun manually, then create the pair relationship. So it  is not used.

PR #524, opened 2018-09-17
Check profile if exist before creating volume

lijuncloud:
Check if the profile exists before creating volume
Issue from : https://github.com/opensds/nbp/issues/127

### DIFF
--- a/pkg/db/drivers/etcd/etcd.go
+++ b/pkg/db/drivers/etcd/etcd.go
@@ -963,8 +963,7 @@ func (c *Client) CreateVolume(ctx *c.Context, vol *model.VolumeSpec) (*model.Vol
 	if err != nil {
 		return nil, err
 	} else if len(profiles) == 0 {
-		log.Error("No profile in db.")
-		return nil, err
+		return nil, errors.New("No profile in db.")
 	}
 
 	vol.TenantId = ctx.TenantId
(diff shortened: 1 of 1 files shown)


pkg/db/drivers/etcd/etcd.go
```diff
@@ -959,6 +959,14 @@ func (c *Client) RemoveExtraProperty(ctx *c.Context, prfID, extraKey string) err
 
 // CreateVolume
 func (c *Client) CreateVolume(ctx *c.Context, vol *model.VolumeSpec) (*model.VolumeSpec, error) {
+	profiles, er := c.ListProfiles(ctx)
```
leonwanghui: Please change it to `err`

pkg/db/drivers/etcd/etcd.go
```diff
@@ -959,6 +959,14 @@ func (c *Client) RemoveExtraProperty(ctx *c.Context, prfID, extraKey string) err
 
 // CreateVolume
 func (c *Client) CreateVolume(ctx *c.Context, vol *model.VolumeSpec) (*model.VolumeSpec, error) {
+	profiles, err := c.ListProfiles(ctx)
+	if err != nil {
+		return nil, err
+	} else if len(profiles) == 0 {
+		log.Error("No profile in db.")
+		return nil, err
```
leonwanghui: You didn't assign any message to err, it would be empty.

PR #525, opened 2018-09-20
Get VolumeType From Cinder instead of Configuration file

lijuncloud:


leonwanghui:
@xing-yang Please take a look

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -412,20 +429,41 @@ func (d *Driver) DeleteSnapshot(req *pb.DeleteVolumeSnapshotOpts) error {
...
+	url := volumeTypeListURL(client)
+	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
+		return StoragePoolPage{pagination.SinglePageBase(r)}
+	})
+}
+
 // ListPools
 func (d *Driver) ListPools() ([]*model.StoragePoolSpec, error) {
-	data, err := ioutil.ReadFile(defaultConfPath)
+	volRes, err := ListVolumeTypes(d.blockStoragev2).AllPages()
```
xing-yang: Was the pool info never saved in OpenSDS database before?

lijuncloud: The created pool will be saved to the database, Is there any question with the logic of the code?

lijuncloud: > Was the pool info never saved in OpenSDS database before?
The created pool will be saved to the database, Is there any question with the logic of the code?

xing-yang: Can you submit a cinder.yaml file?  I don't see it in the repo.  If you don't read from the config file any more, how should the config file be changed?

xing-yang: Can you show me what are the results of "cinder type-list" and "cinder type-show xxxx" on your setup?  schedulerstats only list pools with pool names, but there are no volume types in that list.  
I don't see how you can find a match this way.

xing-yang:
I'm still not sure if volume_type filter for get_pools is supported by GopherCloud yet.  If you look at this PR, volume_type filter will be removed if Cinder API microversion is lower than 3.35.  In the code, I see that blockstorage V2 is used.

contrib/drivers/openstack/cinder/cinder.go
```diff
@@ -170,6 +186,7 @@ func (d *Driver) CreateVolume(req *pb.CreateVolumeOpts) (*model.VolumeSpec, erro
 		Name:        req.GetName(),
 		Description: req.GetDescription(),
 		Size:        int(req.GetSize()),
+		VolumeType:  req.GetPoolName(),
```
xing-yang: Can you submit a separate PR to just pass in the VolumeType specified by user?  Talk to @leonwanghui about it.  Thanks.

lijuncloud: > Can you submit a separate PR to just pass in the VolumeType specified by user? Talk to @leonwanghui about it. Thanks.

OK

skdwriting:
@xing-yang @leonwanghui this looks to be a very old PR. Can we close this PR?

xing-yang:
This work is not complete, but I think we should automatically close stale PRs after a certain period of time.  Authors of the PRs can always restore it and work on it again.

PR #605, opened 2019-02-02
Fix bugs HTTPS support for hotpot

lasshpanzer:
There are some bugs in HTTPS support for hotpot.

Since password encryption is still in development process, so delete the password decryption function when server use password to get token from keystone, user will user original password.


leonwanghui:
Hi @xing-yang, please take a look how to roll back this patch to the master branch before publishing the release, thanks!

wisererik:
please rebase it

xing-yang:
I think @lasshpanzer needs to do a rebase.

PR #612, opened 2019-02-13
Pwd encrypter

jackhaibo:
Fix parsing yaml file bug in pwdEncrypter.go.


jackhaibo:
> s/pwdEncrypter.go/pwdencypter.go
> The file name should not contain upper case character.

Already modified.

PR #674, opened 2019-04-19
Volume status error

### LINKED ISSUE #641: Some errors with status machine

> /kind bug

**What happened**:
Currently when user creates a volume, api-server will create a volume object in database and update the status to `creating`, then controller will handler the rest and finally update the status to `available` or `error` based on the return value. But if some errors occur when apiserver connects controller, the operation would be abandoned and pull a error, but from user's view he can't find any problem and the status of volume would be set `creating` forever.

Change volume status to `error`, rather than `creating`.

* Install opensds cluster
* Stop osdslet service
* Create a volume
* Check the status of volume

@xing-yang @wisererik 

**Environment**:
- Hotpot(release/branch) version: development branch
- OS (e.g. from /etc/os-release): Ubuntu 16.04
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Click2Cloud-Gamma:

**Which issue this PR fixes** *(optional, in `fixes #641` format, will close that issue when PR gets merged)*: fixes #

Is this a BUG REPORT or FEATURE REQUEST?:

/kind bug

What happened:
Currently when user creates a volume, api-server will create a volume object in database and update the status to creating, then controller will handler the rest and finally update the status to available or error based on the return value. But if some errors occur when apiserver connects controller, the operation would be abandoned and pull a error, but from user's view he can't find any problem and the status of volume would be set creating forever.

What you expected to happen:
Change volume status to error, rather than creating.

With this PR the status of Volume creation will become `error` if osdslet controller is not working. Earlier it was showing `creating` status ONLY.

### DIFF
--- a/pkg/api/db.go
+++ b/pkg/api/db.go
@@ -71,8 +71,6 @@ func CreateVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) (*model.VolumeSpe
 	in.Status = model.VolumeCreating
 	// Store the volume data into database.
 	return db.C.CreateVolume(ctx, in)
-	//result, err := db.C.CreateVolume(ctx, in)
-
 }
 
 func CreateVolumeError(ctx *c.Context, in *model.VolumeSpec) {
@@ -86,7 +84,6 @@ func CreateVolumeError(ctx *c.Context, in *model.VolumeSpec) {
 	}
 
 }
-
 // DeleteVolumeDBEntry just modifies the state of the volume to be deleting in
 // the DB, the real deletion operation would be executed in another new thread.
 func DeleteVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) error {
@@ -96,18 +93,6 @@ func DeleteVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) error {
 		errMsg := fmt.Sprintf("only the volume with the status available, error, error_deleting, error_extending can be deleted, the volume status is %s", in.Status)
 		log.Error(errMsg)
 		return errors.New(errMsg)
-	}
-
-	// If profileId or poolId of the volume doesn't exist, it would mean that
-	// the volume provisioning operation failed before the create method in
-	// storage driver was called, therefore the volume entry should be deleted
-	// from db directly.
-	if in.ProfileId == "" || in.PoolId == "" {
-		if err := db.C.DeleteVolume(ctx, in.Id); err != nil {
-			log.Error("when delete volume in db:", err)
-			return err
-		}
-		return nil
 	}
 
 	snaps, err := db.C.ListSnapshotsByVolumeId(ctx, in.Id)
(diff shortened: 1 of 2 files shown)


pkg/api/db.go
```diff
@@ -73,6 +73,17 @@ func CreateVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) (*model.VolumeSpe
 	return db.C.CreateVolume(ctx, in)
 }
 
+func CreateVolumeError(ctx *c.Context, in *model.VolumeSpec) {
+	var errMsg = "Not able to connect to controller"
```
leonwanghui: The first letter should be lower for error message.

pkg/api/db.go
```diff
@@ -73,6 +73,17 @@ func CreateVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) (*model.VolumeSpe
...
+	log.Error(errMsg)
```
leonwanghui: I think `log.Error("fail to connect to controller")` would be better.

pkg/api/volume.go
```diff
@@ -117,7 +117,8 @@ func (v *VolumePortal) CreateVolume() {
 		Context:           ctx.ToJson(),
 	}
 	if _, err = v.CtrClient.CreateVolume(context.Background(), opt); err != nil {
-		log.Error("create volume failed in controller service:", err)
+		CreateVolumeError(ctx, &volume)
+		log.Error("create volume failed in API-server service:", err)
```
leonwanghui: I am wondering if it's better to move this log message before `CreateVolumeError()` method.

leonwanghui:
@Click2Cloud-Gamma Please notice that this PR could be impacted by #676

Click2Cloud-Gamma:
> @Click2Cloud-Gamma Please notice that this PR could be impacted by #676

Yes, as this changed files are placed at different path. I will generate new PR after modifying required changes. Thanks!

pkg/api/volume.go
```diff
@@ -117,7 +117,8 @@ func (v *VolumePortal) CreateVolume() {
 		Context:           ctx.ToJson(),
 	}
 	if _, err = v.CtrClient.CreateVolume(context.Background(), opt); err != nil {
-		log.Error("create volume failed in controller service:", err)
+		CreateVolumeError(ctx, &volume)
```
rhsakarpos: Would it be better to rename this method to 'SetVolumeStatusErrored' or something such...since essentially, it is updating the status of the volume to errorred ?

pkg/api/db.go
```diff
@@ -73,6 +73,17 @@ func CreateVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) (*model.VolumeSpe
...
+	in.UserId = ctx.UserId
+	in.Status = model.VolumeError
```
rhsakarpos: IMHO, it is better we define another status here called ```errorCreating```. This is the actual error right, some error during creating of the volume, which is different from the ```volumeError``` that we already have.

pkg/api/db.go
```diff
@@ -73,6 +73,17 @@ func CreateVolumeDBEntry(ctx *c.Context, in *model.VolumeSpec) (*model.VolumeSpe
 	return db.C.CreateVolume(ctx, in)
 }
 
+func CreateVolumeError(ctx *c.Context, in *model.VolumeSpec) {
```
kumarashit: Can we make this function generic for any Volume Error ops? AFAIU, it's not specific to create volume. VolumeErrorHandle() ?

PR #676, opened 2019-04-19
Refactor API code to make more modular

stmcginnis:
This refactors the API code to follow the beego recommendations and to make it more modular. This restruture should make things a little more manageable as we add file and object support, and it is needed to be able to support alternative northbound APIs (Swordfish) for accessing the some functionality but routed through a different endpoint.

This should not change any existing functionality, so all client code should still work as expected.


pkg/api/routers/router.go
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2017 Huawei Technologies Co., Ltd. All Rights Reserved.
```
leonwanghui: Do we need to change copyright time to 2019?

leonwanghui:
@stmcginnis I think it's a very good job!

pkg/api/routers/router.go
stmcginnis: I've seen mixed advice about whether it really matters or not, but since I did update it elsewhere and missed doing it here, I do think I should change that to be consistent.

PR #725, opened 2019-05-09
Fix a bug in create profile

leonwanghui:
Signed-off-by: leonwanghui <wanghui71leon@gmail.com>

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/controllers/profile.go
+++ b/pkg/api/controllers/profile.go
@@ -22,6 +22,7 @@ package controllers
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/opensds/opensds/pkg/api/policy"
 	c "github.com/opensds/opensds/pkg/context"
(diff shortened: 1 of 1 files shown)


pkg/api/controllers/profile.go
```diff
@@ -54,38 +55,40 @@ func (p *ProfilePortal) CreateProfile() {
...
-
-	capabilityflag := true
-	if profile.StorageType == constants.File {
-		if pp := profile.ProvisioningProperties; !pp.IsEmpty() {
+	stype := strings.ToLower(profile.StorageType)
+	switch stype {
+	case constants.Block:
+		break
+	case constants.File:
+		if pp := profile.ProvisioningProperties; pp.IsEmpty() {
```
xing-yang: Is this required?  We should support default ones if this is not defined. Also admin can define custom properties if pre-defined ones are not used.

Shruthi-1MN: It is to check the StorageAccessCapability, if we won't follow the order then it will consider it as custom property so just to follow the order of properties specification i have added ProvisioningProperties and datastorage. Those are all not pre-defined. Admin still can add custom properties.

PR #727, opened 2019-05-09
Telemetry:Metric Driver mock test (UT)

NajmudheenCT:
Mock test for lvm metric driver .
Collected sample output from host and using this data as the cli output to validate.

Earlier test code was just to call driver methods with no mock values, with which we can't validate  actual vs expected results. Now we can validate all values except timestamp( timesamp is generated by driver collection)


NajmudheenCT:
@xing-yang @skdwriting @wisererik @leonwanghui @rhsprakash

PR #736, opened 2019-05-14
GET API to retrieve Grafana and Alert manager URLs

rhsakarpos:
This PR adds a GET API, which lets the Dashboard retrieve the redirect-to URLs for Grafana and AlertManager components and render the redirect functionality using these URLs

NA

GET 
```
http://localhost:50040/v1beta/e93b4c0934da416eb9c8d120c5d04d96/metrics/urls
```
Response
```
[
    {
        "name": "Grafana",
        "url": "http://localhost:3000"
    },
    {
        "name": "Alert Manager",
        "url": "http://localhost:9093"
    }
]
```
**Release note**:

```
These values will be added to the opensds.conf file after installation. Have mailed the details to Pengyi to implement the same.
```


rhsakarpos:
@skdwriting @wisererik @xing-yang @leonwanghui @NajmudheenCT @anvithks Please review

pkg/api/controllers/metrics.go
```diff
@@ -280,3 +280,27 @@ func (m *MetricsPortal) CollectMetrics() {
 
 	return
 }
+
+func (m *MetricsPortal) GetUrls() {
+	if !policy.Authorize(m.Ctx, "metrics:urls") {
+		return
```
skdwriting: The error info?

rhsakarpos: Checked, in case the authorization fails, we return the status code **403 Forbidden** with the message 
```
{"code":403,"message":"operation is not permitted"}
```

PR #769, opened 2019-05-22
Add Fileshare Api Specs in swagger 

Shruthi-1MN:
Add Fileshare Api Specs in swagger 

**Release note**:

```release-note
```

### DIFF
--- a/openapi-spec/swagger.yaml
+++ b/openapi-spec/swagger.yaml
@@ -153,7 +153,7 @@ paths:
           description: OK
           schema:
             type: array
-            items: 
+            items:
               type: string
         '401':
           $ref: '#/responses/HTTPStatus401'
@@ -331,11 +331,17 @@ paths:
         - Profiles
       description: Updates a profile.
       parameters:
-        - name: body
+        - name: profile
           in: body
-          required: true
           schema:
-            $ref: '#/definitions/ProfileSpec'
+            type: object
+            required:
+              - name
+            properties:
+              name:
+                type: string
+              description:
+                type: string
       responses:
         '200':
           description: OK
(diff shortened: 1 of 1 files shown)


Shruthi-1MN:
> I don't get it why you remove so many information.
I copied from swagger editor so it changed all those, now i corrected the things.

openapi-spec/swagger.yaml
```diff
@@ -1287,6 +1287,296 @@ paths:
...
+            $ref: '#/definitions/FileShareSnapshotSpec'
+        '400':
+          $ref: '#/responses/HTTPStatus400'
+        '401':
+          $ref: '#/responses/HTTPStatus401'
+        '403':
+          $ref: '#/responses/HTTPStatus403'
+        '500':
+          $ref: '#/responses/HTTPStatus500'
+  '/v1beta/{tenantId}/file/snapshots/{fileshareId}':
```
leonwanghui: File share could have multiple snapshot at the same time, so please use `shareSnapshotId` here.

Shruthi-1MN: Yes file share supports multiple snapshots, and they can create using existing fileshareId so source resource is fileshareId, how it will be suitable if we replace fileshareId with shareSnapshotId.

Shruthi-1MN: You mean should i have to add one more api for creating snapshotId for existing snapshot?

leonwanghui: No, just replace `shareSnapshotId` with `fileshareId`

Shruthi-1MN: ok i got it. to update, get, delete i should add sharesnapshotId. Done.

openapi-spec/swagger.yaml
```diff
@@ -2044,8 +2497,8 @@ parameters:
     required: true
     description: The UUID of the volume attachment.
     type: string
-  snapshotId:
-    name: snapshotId
+  shareSnapshotId:
```
leonwanghui: snapshotId is volume snapshot id, you need to add another variable.

Shruthi-1MN: Done

Issue #789, opened 2019-05-28
Telemetry: Naming format wrong for metrics pushed through pushgateway


NajmudheenCT:
**Is this a BUG REPORT or FEATURE REQUEST?**:

 /kind bug

**What happened**:
When pushgateway is used as push mechanism in metrics controller , the name formed by controller is not in consistent with lvm_exporter .
now name is coming like 
read_throughput{device="loop28",instance="",job="push_gateway",metricname="read_throughput"} 0
**What you expected to happen**:
expected name 
lvm_volume_read_throughput_kbs{device="loop28"} 0
**How to reproduce it (as minimally and precisely as possible)**:
call POST metrics API using any client.

PR #793, opened 2019-05-29
fix #788 and #789

### LINKED ISSUE #788: Telemetry: metric name ends with "_" when AggrType is nil

 /kind bug

**What happened**:
When node_exporter is configured as push mechanism , the metric name created by Prometheus adaptor in hotpot controller ends with _ . this is because  AggrType of lvm metrics are nil.
if a field is nil, do not append  "_"  
call POST metric API , using node exporter as push mechanism

rhsakarpos:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #788 and #789 

**Release note**:

```
None
```


rhsakarpos:
@NajmudheenCT @leonwanghui @skdwriting @xing-yang please review.

PR #794, opened 2019-05-29
Adding profile to CreateFileShareAcl and snapshot issue

PravinRanjan10:
It add Profile to CreateFileShareAcl controller and fix issue related to snapshot metadata.
It also has dock and controller logic for deleteaccessAPI

**Release note**:

```release-note
```

### DIFF
--- a/pkg/model/proto/model.proto
+++ b/pkg/model/proto/model.proto
@@ -166,6 +166,9 @@ service FileShareController {
     // Create a file share Acl
     rpc CreateFileShareAcl (CreateFileShareAclOpts) returns (GenericResponse){}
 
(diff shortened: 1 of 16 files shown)


pkg/model/proto/model.proto
```diff
@@ -831,6 +831,8 @@ message DeleteFileShareSnapshotOpts {
     string context = 4;
     // The Serialized profile
     string profile = 5;
+    // The metadata of the fileshare, optional.
+    map<string, string> metadata = 9;
```
leonwanghui: This should be `6` here.

PravinRanjan10: done

Issue #800, opened 2019-05-31
Telemetry: Controller is not picking urls from config


NajmudheenCT:
**Is this a BUG REPORT or FEATURE REQUEST?**:
metrics/url request is failing in controller as the flag which controller looks for is not set initially.

Solution is to set flag in osdslet.init
flag.StringVar(&CONF.OsdsApiServer.GrafanaUrl, "grafana-url", CONF.OsdsApiServer.GrafanaUrl, "Grafana listen endpoint")
	flag.StringVar(&CONF.OsdsApiServer.AlertMgrUrl, "alertmgr-url", CONF.OsdsApiServer.AlertMgrUrl, "Alert manager listen endpoint")
	

 /kind bug

**What happened**:
getUrl return null
**What you expected to happen**:
get url is expected to return 
[
    {
        "name": "grafana-url",
        "url": "http://localhost:3000"
    },
    {
        "name": "alertmgr-url",
        "url": "http://localhost:9093"
    }
]
**How to reproduce it (as minimally and precisely as possible)**:
GET API http://127.0.0.1:50040/v1beta/e93b4c0934da416eb9c8d120c5d04d96/metrics/urls

PR #826, opened 2019-06-07
Telemetry : Fix for https://github.com/opensds/opensds/issues/800

NajmudheenCT:
Fix for https://github.com/opensds/opensds/issues/800
metrics/url request is failing in controller as the flag which controller looks for is not set initially.

Solution is to set flag in osdslet.init
**Which issue this PR fixes** 
https://github.com/opensds/opensds/issues/800


NajmudheenCT:
@leonwanghui @xing-yang @skdwriting @wisererik

PR #838, opened 2019-06-11
Telemetry: Ceph volume metrics with updated Ceph Metrics drivers' unit test case

Click2Cloud-Gamma:
With this PR, we can collect the following metrics related to volume in the Ceph cluster:
1. ceph_volume_name
2. ceph_volume_size_bytes
3. ceph_volume_objects
4. ceph_volume_object_size_bytes


leonwanghui:
@wisererik PTAL

Issue #846, opened 2019-06-12
Telemetry: lvm_exporter Guauge name for disk metrics is not consistent with Driver


NajmudheenCT:
Telemetry: lvm_exporter Guauge name for disk metrics is not consistent with Driver
for disk metrics , instead of having name like 
lvm_disk_iops_tps

exporter name is lvm_volume_iops_tps
 /kind bug
> /kind feature

PR #848, opened 2019-06-12
Telemtery: Fix for https://github.com/opensds/opensds/issues/846

NajmudheenCT:
Telemetry: lvm_exporter Guauge name for disk metrics is not consistent with Driver
for disk metrics , 
exporter puts volume instead of disk in name 

**Which issue this PR fixes** 
https://github.com/opensds/opensds/issues/846


NajmudheenCT:
please review @leonwanghui @wisererik @skdwriting

PR #852, opened 2019-06-13
hotpot connector for NFS

jackhaibo:
Connector of file share `NFS` protocol used by sushi.

### DIFF
--- a/contrib/connector/connector.go
+++ b/contrib/connector/connector.go
@@ -35,6 +35,7 @@ const (
 
 	NvmeofDriver = config.NVMEOFProtocol
 	Nqn          = "nqn"
+	NFSDriver    = "NFS"
 )
 
 // Connector implementation
--- a/contrib/connector/nfs/helper.go
+++ b/contrib/connector/nfs/helper.go
@@ -1,0 +1,108 @@
+// Copyright 2019 The OpenSDS Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


leonwanghui:
@wisererik @xxwjj PTAL

contrib/connector/connector.go
```diff
@@ -35,6 +35,7 @@ const (
 
 	NvmeofDriver = config.NVMEOFProtocol
 	Nqn          = "nqn"
+	NFSDriver    = "NFS"
```
leonwanghui: Please use config.NFSProtocol instead.

contrib/connector/nfs/helper.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/mitchellh/mapstructure"
+	"github.com/opensds/opensds/contrib/connector"
+)
+
+import (
+	"fmt"
+)
+
+type NFSConnectorInfo struct {
+	ExportLocations []string `mapstructure:"ExportLocations"`
```
leonwanghui: is here ExportLocations or exportLocations?

contrib/connector/connector.go
jackhaibo: Done.

contrib/connector/nfs/helper.go
jackhaibo: Change it to exportLocations.

PR #877, opened 2019-06-15
Fix bugs relate on oceanstor file share driver

jackhaibo:
Fix bugs relate on file share.

### DIFF
--- a/contrib/drivers/filesharedrivers/oceanstor/oceanstor.go
+++ b/contrib/drivers/filesharedrivers/oceanstor/oceanstor.go
@@ -27,6 +27,7 @@ import (
 	. "github.com/opensds/opensds/contrib/drivers/utils/config"
 	model "github.com/opensds/opensds/pkg/model"
 	pb "github.com/opensds/opensds/pkg/model/proto"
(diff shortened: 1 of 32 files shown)


contrib/drivers/filesharedrivers/oceanstor/oceanstor.go
```diff
@@ -473,34 +474,26 @@ func (d *Driver) DeleteFileShareSnapshot(opt *pb.DeleteFileShareSnapshotOpts) er
...
+	} else {
+		accessLevel = "ro"
 	}
 
-	if accessLevel == "r" {
-		accessLevel += "o"
+	if !utils.Contained(AccessLevelRead, accessLevels) || !utils.Contained(AccessLevelWrite, accessLevels) {
+		return "", errors.New("only read only or read write access level are supported")
+	} else {
+		accessLevel = "rw"
```
wisererik: the conditional codes is not right

skdwriting: only write case will pass? please check the conditional code.

PR #883, opened 2019-06-17
Add accessprotocol to default profile for fileshare

Shruthi-1MN:

**Release note**:

```release-note
```


leonwanghui:
Could we set some fields by default in the source? eg. when creating profile, if accessProtocol is empty, set it `nfs`, and if accessCapabilities is empty, set it [Read, Write, Execute]

Shruthi-1MN:
> Could we set some fields by default in the source? eg. when creating profile, if accessProtocol is empty, set it `nfs`, and if accessCapabilities is empty, set it [Read, Write, Execute]
 Nice idea. Thanks! Added in source code as default values.

PR #888, opened 2019-06-17
[DO NOT merge] Thin opensds

leonwanghui:
**What this PR does / why we need it**:
This PR based on #696 is just to show what's changed in [thin-opensds](https://github.com/opensds/opensds/tree/thin-opensds) branch compared to latest code, please do NOT merge this PR.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #894, opened 2019-06-18
Keep ProvisionDockServer service consistent with ProvisionControllerServer service

leonwanghui:
Signed-off-by: leonwanghui <wanghui71leon@gmail.com>

This PR is proposed to add `GetMetrics` and `GetUrls` method in ProvisionDockServer service so as to keep consistent with ProvisionControllerServer service. Besides, some field validation code is also added in apiserver module.

**Release note**:

```release-note
```


leonwanghui:
@xing-yang @wisererik PTAL, thanks!

pkg/api/controllers/profile.go
```diff
@@ -64,7 +64,7 @@ func (p *ProfilePortal) CreateProfile() {
 				profile.ProvisioningProperties.DataStorage.StorageAccessCapability = []string{"Read", "Write", "Execute"}
 			}
 		}
-		if io := pp.IOConnectivity; io.IsEmpty(){
+		if io := pp.IOConnectivity; io.IsEmpty() {
```
skdwriting: just a line position change?! no code change..

pkg/api/util/db.go
```diff
@@ -120,6 +125,12 @@ func CreateFileShareDBEntry(ctx *c.Context, in *model.FileShareSpec) (*model.Fil
 	if in.Id == "" {
 		in.Id = uuid.NewV4().String()
 	}
+	// validate profileId
+	if in.ProfileId == "" {
```
skdwriting: really need this duplicate check code? cannot be avoided?

pkg/dock/dock.go
```diff
@@ -585,6 +584,18 @@ func (ds *dockServer) DeleteFileShareSnapshot(ctx context.Context, opt *pb.Delet
...
 		return pb.GenericResponseError(err), err
 	}
-	// TODO: maybe need to update status in DB.
+
 	return pb.GenericResponseResult(nil), nil
 }
+
+// GetMetrics method is only defined to make ProvisioinDock service consistent with
+// ProvisionController service, so this method is not allowed to be called.
+func (ds *dockServer) GetMetrics(context.Context, *pb.GetMetricsOpts) (*pb.GenericResponse, error) {
```
skdwriting: Why do we need these two functions just to send the message that it is not implemented?

pkg/model/proto/model.proto
```diff
@@ -147,6 +147,12 @@ service ProvisionDock {
 
     // Collect metrics from metrics driver
     rpc CollectMetrics (CollectMetricsOpts) returns (GenericResponse){}
+
+    // Get metrics from Prometheus
+    rpc GetMetrics (GetMetricsOpts) returns (GenericResponse){}
```
skdwriting: purpose?! same for GetUrls

pkg/api/controllers/profile.go
leonwanghui: That's the impact after running goimports command.

pkg/dock/dock.go
leonwanghui: If you look at [comments](https://github.com/opensds/opensds/pull/888#discussion_r294332438) from xing, we came to conclusion that we should keep thin-opensds as consistent as possible with master branch.

pkg/model/proto/model.proto
leonwanghui: Please see the comment above.

xing-yang:
Please rebase.

leonwanghui:
@xing-yang Please take a review, thanks!

xing-yang:
@leonwanghui The changes look good to me.  Can you please squash your commits into one?  Thanks.

PR #912, opened 2019-06-20
Solve some volume group update and delete errors.

jackhaibo:
Solve some volume group update and delete errors.

### DIFF
--- a/contrib/drivers/lvm/targets/iscsi.go
+++ b/contrib/drivers/lvm/targets/iscsi.go
@@ -86,8 +86,9 @@ func (t *tgtTarget) getTgtConfPath(volId string) string {
 type configMap map[string][]string
 
 func (t *tgtTarget) CreateISCSITarget(volId, tgtIqn, path, hostIp, initiator string, chapAuth []string) error {
-	if hostIp == "" {
-		return errors.New("create ISCSI target failed, host ip cannot be empty")
+	// Multi-attach require a specific ip
+	if hostIp == "" || hostIp == "ALL" {
+		return errors.New("create ISCSI target failed: host ip cannot be empty")
 	}
 
 	if exist, _ := utils.PathExists(t.TgtConfDir); !exist {
--- a/contrib/drivers/lvm/targets/iscsi.go
+++ b/contrib/drivers/lvm/targets/iscsi.go
@@ -89,12 +89,16 @@ type configMap map[string][]string
 func (t *tgtTarget) CreateISCSITarget(volId, tgtIqn, path, hostIp, initiator string, chapAuth []string) error {
 	// Multi-attach require a specific ip
 	if hostIp == "" || hostIp == "ALL" {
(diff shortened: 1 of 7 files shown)


leonwanghui:
@wisererik @xxwjj PTAL

contrib/drivers/lvm/targets/iscsi.go
```diff
@@ -86,8 +86,9 @@ func (t *tgtTarget) getTgtConfPath(volId string) string {
 type configMap map[string][]string
 
 func (t *tgtTarget) CreateISCSITarget(volId, tgtIqn, path, hostIp, initiator string, chapAuth []string) error {
-	if hostIp == "" {
-		return errors.New("create ISCSI target failed, host ip cannot be empty")
+	// Multi-attach require a specific ip
+	if hostIp == "" || hostIp == "ALL" {
+		return errors.New("create ISCSI target failed: host ip cannot be empty")
```
xxwjj: the error is not corrct.

leonwanghui: Besides, some log msg may be required here.

jackhaibo: done.

contrib/drivers/lvm/targets/iscsi.go
```diff
@@ -86,8 +87,14 @@ func (t *tgtTarget) getTgtConfPath(volId string) string {
...
-	if hostIp == "" {
-		return errors.New("create ISCSI target failed, host ip cannot be empty")
+	// Multi-attach require a specific ip
+	if hostIp == "" || hostIp == "ALL" {
+		return fmt.Errorf("create ISCSI target failed: host ip %s cannot be empty or ALL, iscsi only allows specific ip access, not all", hostIp)
+	}
+
+	result := net.ParseIP(hostIp)
+	if result == nil {
+		return fmt.Errorf("%s is not a valid ip. Please give the proper ip", hostIp)
```
leonwanghui: %s is not a valid ip, please give the proper ip

leonwanghui: Also require log here.

pkg/api/controllers/volumeGroup.go
```diff
@@ -127,21 +127,38 @@ func (v *VolumeGroupPortal) UpdateVolumeGroup() {
...
 	if err != nil {
 		errMsg := fmt.Sprintf("marshal volume group updated result failed: %s", err.Error())
 		v.ErrorHandle(model.ErrorInternalServer, errMsg)
 		return
 	}
+
 	v.SuccessHandle(StatusAccepted, body)
 
+	// No more values in group need to be updated
+	if addVolumes == nil && removeVolumes == nil {
```
leonwanghui: I prefer using `len(addVolumes) == 0 && len(removeVolumes) == 0` here.

PR #948, opened 2019-07-08
File Share ACL create and delete services updating file share status are corrected - Fix #930

### LINKED ISSUE #930: Dashboard-FileShare: After Some Access rules creation, FileShare status becoming error from available

>
> /kind bug
**What happened**:
1. Create File Share and ensure its status is Available
2. create Access Rules more than 10 ( either in same Popup or seperately)
3. No response in dashboard after clicking OK Button
4. In Browser Response, we can see 400 error : {"code":400,"message":"createFileshareAcldbentry failed: only the status of file share is available, the acl can be created"}
5. Click cancel of popu and go and refresh File share page. File share status changed to ERROR from Available 
1. Should create all the rules
2. If error, specific error should display in UI instead of showing same page without any error
3. Respose error statement also ambigious.. should be simple and understandable
FileShareDashBoard_After some rules creation, Fileshare becoming error

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Shruthi-1MN:
Problem statement:
==============
Create file share acl code was updating file share status when it fails to create and deletion of acl as well.
There was not status field in Fileshareaclspec to update it's status when it is successful or fails.
So #794 PR is updating fileshare status when acl creation fails, it was causing fatal error so issue has been raised #930 

Solution/Fix:
==========
Added  status field in Fileshareaclspec to tract it's status of creation and failure.
Whenever ACL creation fails to create, it will be update file share acl status.

**Release note**:

```release-note
```

### DIFF
--- a/pkg/api/util/db.go
+++ b/pkg/api/util/db.go
@@ -109,19 +109,36 @@ func CreateFileShareAclDBEntry(ctx *c.Context, in *model.FileShareAclSpec) (*mod
 		return nil, errors.New(errMsg)
 	}
 
-	// Store the fileshare meadata into database.
 	return db.C.CreateFileShareAcl(ctx, in)
 }
 
 func DeleteFileShareAclDBEntry(ctx *c.Context, in *model.FileShareAclSpec) error {
-	// If fileshare id is invalid, it would mean that fileshare snapshot
+	// If fileshare id is invalid, it would mean that fileshare acl
 	// creation failed before the create method in storage driver was
 	// called, and delete its db entry directly.
-	if _, err := db.C.GetFileShare(ctx, in.FileShareId); err != nil {
+	validStatus := []string{model.FileShareAclAvailable, model.FileShareAclError,
+		model.FileShareAclErrorDeleting}
+	if !utils.Contained(in.Status, validStatus) {
+		errMsg := fmt.Sprintf("only the file share acl with the status available, error, error_deleting can be deleted, the fileshare status is %s", in.Status)
+		log.Error(errMsg)
+		return errors.New(errMsg)
+	}
+
+	// If fileshare id is invalid, it would mean that file share acl creation failed before the create method
+	// in storage driver was called, and delete its db entry directly.
+	_, err := db.C.GetFileShare(ctx, in.FileShareId)
+	if err != nil {
 		if err := db.C.DeleteFileShareAcl(ctx, in.Id); err != nil {
-			log.Error("when delete fileshare acl in db:", err)
+			log.Error("failed delete fileshare acl in db:", err)
 			return err
 		}
+		return nil
+	}
+
+	in.Status = model.FileShareAclDeleting
+	_, err = db.C.UpdateFileShareAcl(ctx, in)
+	if err != nil {
+		return err
 	}
 	return nil
 }
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -953,13 +953,13 @@ func (c *Controller) CreateFileShareAcl(contx context.Context, opt *pb.CreateFil
 
 	fileshare, err := db.C.GetFileShare(ctx, opt.FileshareId)
 	if err != nil {
-		db.UpdateFileShareStatus(ctx, db.C, opt.Id, model.FileShareError)
+		db.UpdateFileShareAclStatus(ctx, db.C, opt.Id, model.FileShareAclError)
 		return pb.GenericResponseError(err), err
 	}
 
 	dockInfo, err := db.C.GetDockByPoolId(ctx, fileshare.PoolId)
 	if err != nil {
-		db.UpdateFileShareStatus(ctx, db.C, opt.Id, model.FileShareError)
+		db.UpdateFileShareAclStatus(ctx, db.C, opt.Id, model.FileShareAclError)
 		log.Error("when search supported dock resource:", err.Error())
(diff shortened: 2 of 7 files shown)


PravinRanjan10:
One question, why we need status for fileshare ACL??. How it solve the #930.

Shruthi-1MN:
> One question, why we need status for fileshare ACL??. How it solve the #930.
Please check the old code, if acl doesn't create successfully then your were updating fileshare status. Actually you should update file share acl status. 
if err != nil {
		// Change the status of the file share acl to error when the creation faild
		defer db.UpdateFileShareStatus(ctx, db.C, fileshare.Id, model.FileShareError)
		log.Error("when create file share acl:", err.Error())
		return pb.GenericResponseError(err), err
	}

leonwanghui:
@Shruthi-1MN One question is if the acl has no status, why do we need to update the status of fileshare if acl creation fails?

Shruthi-1MN:
> @Shruthi-1MN One question is if the acl has no status, why do we need to update the status of fileshare if acl creation fails?
Good question, that what old code was doing so in this PR i have corrected and added status field for acl to update it's status not file share status

testutils/collection/data.go
```diff
@@ -229,7 +229,7 @@ var (
 			ExportLocations:  []string{"192.168.100.101"},
 		},
 	}
-
+
```
leonwanghui: Please remove the blank spaces.

pkg/model/proto/model.proto
```diff
@@ -758,6 +758,8 @@ message DeleteFileShareAclOpts {
     string Name = 10;
     // The metadata of the file share, optional.
     map<string, string> metadata = 11;
+    // The status
+    string status = 12;
```
leonwanghui: Do we need to persist status when sending gRPC request?

pkg/model/fileshare.go
```diff
@@ -39,6 +39,10 @@ type FileShareAclSpec struct {
 	// The description of the fileshare acl.
 	Description string `json:"description,omitempty"`
 
+	// The status of the fileshareacl.
+	// One of: "available", "error" etc.
+	Status string `json:"status,omitempty"`
```
leonwanghui: If you want to add the Status field, please also change swagger.yaml

Shruthi-1MN: ok i will raise another PR to correct swagger.yaml

pkg/model/proto/model.proto
```diff
@@ -784,6 +784,8 @@ message CreateFileShareAclOpts {
...
```
leonwanghui: Please remove status field, it's not required here.

pkg/controller/controller.go
```diff
@@ -987,10 +987,13 @@ func (c *Controller) DeleteFileShareAcl(contx context.Context, opt *pb.DeleteFil
 
 	fileshare, err := db.C.GetFileShare(ctx, opt.FileshareId)
 	if err != nil {
+		defer db.UpdateFileShareAclStatus(ctx, db.C, fileshare.Id, model.FileShareAclErrorDeleting)
+		log.Error("when delete file share acl:", err.Error())
 		return pb.GenericResponseError(err), err
 	}
 	dockInfo, err := db.C.GetDockByPoolId(ctx, fileshare.PoolId)
 	if err != nil {
+		defer db.UpdateFileShareAclStatus(ctx, db.C, fileshare.Id, model.FileShareAclErrorDeleting)
```
leonwanghui: It should not be fileshare.Id here, please change it to acl id.

pkg/api/util/db.go
```diff
@@ -108,19 +108,37 @@ func CreateFileShareAclDBEntry(ctx *c.Context, in *model.FileShareAclSpec) (*mod
 		return nil, errors.New(errMsg)
 	}
 
-	// Store the fileshare meadata into database.
+	in.Status = model.FileShareAclAvailable
```
leonwanghui: Why this status is changed into `available`?

pkg/api/util/db.go
```diff
@@ -108,19 +108,37 @@ func CreateFileShareAclDBEntry(ctx *c.Context, in *model.FileShareAclSpec) (*mod
...
 	return db.C.CreateFileShareAcl(ctx, in)
 }
 
 func DeleteFileShareAclDBEntry(ctx *c.Context, in *model.FileShareAclSpec) error {
```
leonwanghui: Please also add some test cases for failing to delete invalid file share acl.

pkg/controller/controller.go
```diff
@@ -970,7 +970,7 @@ func (c *Controller) CreateFileShareAcl(contx context.Context, opt *pb.CreateFil
 	result, err := c.fileshareController.CreateFileShareAcl((*pb.CreateFileShareAclOpts)(opt))
 	if err != nil {
 		// Change the status of the file share acl to error when the creation faild
-		defer db.UpdateFileShareStatus(ctx, db.C, fileshare.Id, model.FileShareError)
+		defer db.UpdateFileShareAclStatus(ctx, db.C, opt.Id, model.FileShareAclError)
```
leonwanghui: You need to make the file share acl status to `available` if everything works fine.

Shruthi-1MN: 139 has update status in pkg/api/util/db.go file

PR #958, opened 2019-07-10
sample connector for csi-test

jackhaibo:
Add sample connector and interface the mounter to support csi-test of sushi.
1. Add sample connector
2. Interface the mounter


contrib/connector/common.go
```diff
@@ -22,6 +22,22 @@ import (
...
+	Mount(device, mountpoint, fsType string, mountFlags []string) error
+	Umount(mountpoint string) error
+	IsMounted(target string) (bool, error)
+	GetHostName() (string, error)
+	GetHostIP() string
+}
+
+type CommonMounter struct{}
+
+func GetCommonMounter() Mounter {
```
leonwanghui: s/GetCommonMounter/NewMounter

contrib/connector/common.go
```diff
@@ -22,6 +22,22 @@ import (
...
+	GetFSType(device string) (string, error)
+	Format(device string, fsType string) error
+	Mount(device, mountpoint, fsType string, mountFlags []string) error
+	Umount(mountpoint string) error
+	IsMounted(target string) (bool, error)
+	GetHostName() (string, error)
+	GetHostIP() string
+}
+
+type CommonMounter struct{}
```
leonwanghui: Maybe it's better to make it private.

contrib/connector/common.go
jackhaibo: `CommonMounter` or `NewMounter`, because sushi mounter is defined as `FakeMounter`

jackhaibo: Done.

contrib/connector/common.go
jackhaibo: Done.

PR #967, opened 2019-07-18
driver: introduce chubaofs fileshare driver

shuoranliu:

This commit introduces the chubaofs FileShareDriver.

N/A

The driver has following functions implemented:

CreateFileShare/DeleteFileShare/ListPools

Note that ListPools is just a stub function since chubaofs does not have such concept as "pool". ChubaoFS is already over-provisioned.

**Release note**:

```release-note

ChubaoFS fileshare driver.

The following functions are implemented for now:
- CreateFileShare/DeleteFileShare
- ListPools

Note that ListPools is just a stub function since chubaofs does not have such concept as "pool". ChubaoFS is already over-provisioned.

The following functions are not implemented yet:
- CreateFileShareAcl/DeleteFileShareAcl
- CreateFileShareSnapshot/DeleteFileShareSnapshot

Note that the access control relies on the default UNIX file permissions, i.e. uid, gid and rwx permissions.

```

### DIFF
--- a/contrib/drivers/filesharedrivers/chubaofs/chubaofs.go
+++ b/contrib/drivers/filesharedrivers/chubaofs/chubaofs.go
@@ -1,0 +1,249 @@
+// Copyright (c) 2019 The OpenSDS Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package chubaofs
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path"
+
+	log "github.com/golang/glog"
+	. "github.com/opensds/opensds/contrib/drivers/utils/config"
+	. "github.com/opensds/opensds/pkg/model"
+	pb "github.com/opensds/opensds/pkg/model/proto"
+	"github.com/opensds/opensds/pkg/utils/config"
+	uuid "github.com/satori/go.uuid"
+)
+
+const (
+	DefaultConfPath = "/etc/opensds/driver/chubaofs.yaml"
+	NamePrefix      = "chubaofs"
+)
+
+const (
+	KMountPoint = "mountPoint"
+	KVolumeName = "volName"
+	KMasterAddr = "masterAddr"
+	KLogDir     = "logDir"
+	KWarnLogDir = "warnLogDir"
+	KLogLevel   = "logLevel"
+	KOwner      = "owner"
+	KProfPort   = "profPort"
+)
+
+const (
+	KClientPath = "clientPath"
+)
+
+const (
+	defaultLogLevel = "error"
+	defaultOwner    = "chubaofs"
+	defaultProfPort = "10094"
+
+	defaultVolumeCapLimit int64 = 1000000
+)
+
+const (
+	clientConfigFileName = "client.json"
+	clientCmdName        = "cfs-client"
+)
+
+type ClusterInfo struct {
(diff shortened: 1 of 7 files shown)


contrib/drivers/filesharedrivers/chubaofs/chubaofs.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func (d *Driver) Setup() error {
+	conf := &Config{}
+	path := config.CONF.OsdsDock.Backends.Chubaofs.ConfigPath
+	if "" == path {
+		path = DefaultConfPath
+	}
+	_, err := Parse(conf, path)
+
```
wisererik: need to handle err here

shuoranliu: thanks, done.

contrib/drivers/filesharedrivers/chubaofs/chubaofs.go
```diff
@@ -0,0 +1,249 @@
...
+	configFiles, fsMntPoints, owner, err := prepareConfigFiles(d, opt)
+	if err != nil {
+		return nil, err
+	}
+
+	/*
+	 * Only the master raft leader can repsonse to create volume requests.
+	 */
+	leader, err := getClusterInfo(d.conf.MasterAddr[0])
```
kumarashit: Is the master address updated by some scheduler/daemon? In case there is new master/leader elected, will it get the updated info?

contrib/drivers/filesharedrivers/chubaofs/util.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}()
+
+	if err = os.MkdirAll(clientWarnLog, os.ModeDir); err != nil {
+		err = errors.New(fmt.Sprintf("chubaofs: failed to create client warn log dir, path: %v , err: %v", clientWarnLog, err))
+		return
+	}
+	defer func() {
+		if err != nil {
+			os.RemoveAll(clientWarnLog)
```
kumarashit: Can we have some logging before removing warnlog, log or config?

contrib/drivers/filesharedrivers/chubaofs/chubaofs.go
shuoranliu: The master address is specified in the driver config file, and will not be changed. So it is either several static ip addresses, or a domain name without ip address predefined. It is according to the on-premise cluster's setup.

contrib/drivers/filesharedrivers/chubaofs/util.go
shuoranliu: Sure, I'll add some warnings here. However, just for clarification, these directories are created during the volume creation process, so upon reaching here, the creation just failed.

shuoranliu:
@kumarashit sorry for the late reply. All the comments are addressed. Thanks.

contrib/drivers/filesharedrivers/chubaofs/chubaofs.go
kumarashit: Ok. Then it's fine. 
I was under the impression that if a master goes down, new leader/master will be elected as per RAFT

kumarashit:
@shuoranliu that's fine.

xing-yang:
Hi @shuoranliu !  Thanks for contributing the ChuboFS driver.  I see that Create/DeleteFileShareAcl are not implemented.  Does that mean file shares created by ChuboFS driver will have a default access control level?  What is that?  Please clarify.

Can you please fill in the release notes section at the top of this PR, describing what functions are implemented and what are not (including what is the default access control level, etc.)?  Thanks.

shuoranliu:
@xing-yang yes, the Create/DeleteFileShareAcl are not implemented, so the access control replys on the default UNIX file permissions, i.e. uid, gid and rwx permissions. This is because the implementation of filesystem ACL is based on "xattr" feature which is not supported by ChubaoFS yet. But we are thinking of adding this feature. Maybe after then we can update this driver and implement Create/DeleteFileShareAcl functions.

I'll add a detailed release note.

shuoranliu:
@xing-yang release note is added, thanks for the review.

xing-yang:
lgtm

PR #977, opened 2019-07-26
CI Files for UT, IT and E2E Test

satya-gorli:
What this PR does / why we need it:
The PR contains files related to Static Checks (goMetalint, goFmt,) and testing (Unit testing, integration testing, e2e test) jobs
Which issue this PR fixes (optional, in fixes #(, fixes #<issue_number>, ...) format, will close that issue when PR gets merged): fixes #
CI files
Special notes for your reviewer:
Please review the CI Files.. to unblock other PRs, GoLint call is not put in Makefile
Release note:
None


osds/verify.sh
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+if ! [ -f "./bin/gometalinter" ]; then echo "Download The files";
+	curl -L https://git.io/vp6lP | sh
+fi
+export PATH=${PATH}:${GOPATH}/bin:${GOPATH}/src/github.com/opensds/opensds/osds/bin
+gometalinter --disable-all --enable=gofmt --enable=misspell --enable=golint --exclude=vendor --exclude=test ./...
```
kumarashit: misspell may report some changes which may not be required right? Also, it won't fail the build right?

wisererik: +1 , same confusion here, I didn't see any check in CI log files

kumarashit:
LGTM. Overall
- Some of the misspell warnings can be misleading. Ex: Suggestion of replacing exportfs with exports.
- There will be lots of comment related warnings exported function which should be followed but herculean task to fix it for existing code. Let's not fail the build for these type of warnings
- Good that it warns for some variable naming like _ should not be there in variable name and naming 'Id' to ID. Leads to good coding

kumarashit:
Please check the branch out-of-date issue

.travis.yml
```diff
@@ -44,9 +46,20 @@ install:
   # Build OpenSDS Controller source code
   - make all
 
-script:
-  - ./install/CI/coverage
-  - ./install/CI/test
+
```
rhsakarpos: how is this supposed to work ? series or parallel ? single failure stops all next steps ? in clouddragon etc., we have

1. ability to define some steps as mandatory, some as optional (to PASS, an optional FAILURE will NOT break the build)
2. ability to enable/disable steps to reduce build time (very useful when close to release)
3. ability to run sequential (predecessor/successor model) or parallel
please check.

osds/verify.sh
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+if ! [ -f "./bin/gometalinter" ]; then echo "Download The files";
+	curl -L https://git.io/vp6lP | sh
+fi
+export PATH=${PATH}:${GOPATH}/bin:${GOPATH}/src/github.com/opensds/opensds/osds/bin
+gometalinter --disable-all --enable=gofmt --enable=misspell --enable=golint --exclude=vendor --exclude=test ./../... > golintoutput.txt
+if [ $? != 0 ]; then
```
rhsakarpos: each of the sub task scripts should return the status (pass/fail with details) to the outside caller, right ?

test/unittest/unittest.sh
```diff
@@ -0,0 +1,44 @@
...
+# Start the openapi-spec validation.
+split_line "Start openapi-spec validation"
+
+wget https://github.com/go-swagger/go-swagger/releases/download/v0.19.0/swagger_linux_amd64
+chmod +x ./swagger_linux_amd64 && ./swagger_linux_amd64 validate --stop-on-error $OPENSDS_DIR/openapi-spec/swagger.yaml
+rm ./swagger_linux_amd64
+
+# Start unit test.
+split_line "Start unit test"
+go test -v github.com/opensds/opensds/osdsctl/... -cover
```
rhsakarpos: is the latest code also pulled always from github ?

osds/verify.sh
satya-gorli: golint output is written to file.. each one we need to analyse and plan during Daito

.travis.yml
satya-gorli: In Travis file different jobs are defined.
if we want specific jobs, we can run by passing job name
and all these jobs run parallel
CIJobs

osds/verify.sh
satya-gorli: if [ $? != 0 ]; then --> this check will return the status
but not to fail build, not returning error on failure... otherwise for all PRs, build will fail as many golint warnings are there

test/unittest/unittest.sh
satya-gorli: Yes.. In Travis CI, it will take latest code and run the builds

osds/verify.sh
skdwriting: good

skdwriting:
Please raise another issue to track the golint issues. thanks

PR #979, opened 2019-08-01
add logs function to standardize the log format

liuqinguestc:
add logs function to standardize the log format

**Release note**:

```release-note
```

### DIFF
--- a/pkg/utils/logs/logs.go
+++ b/pkg/utils/logs/logs.go
@@ -20,6 +20,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/utils/logs/logs.go
```diff
@@ -80,3 +81,24 @@ func InitLogs(LogFlushFrequency time.Duration) {
...
+//timeNow := time.Now().UTC().Format("2006-01-02 15:04:05.000000")
+//LogPrefix: multi-cloud、hotpot、nbp
+//os.Getpid: current process id
+//logLevel: Log level,such as INFO、WARNING、ERROR、FATAL.
+//filename: go file name.
+//line: Line Number.
+//funcname: function name
+//logMsg: log detail Messages
+//errorCode: if logLevel is ERROR ,else errorCode=""
+func GetLogContent(logPrefix string, pid string, logLevel string, fileName string, line string, funcName string, logMsg string, errorCode string) string {
```
kumarashit: From Where this function is getting called? Any common location?

pkg/utils/logs/logs_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestLogContent(t *testing.T) {
+	logPrefix := "hotpot"
+	pid := "11111"
+	logLevel := "ERROR"
+	filename := "test.go"
+	line := "20"
+	funcname := "testFunc"
+	logMsg := "Hello, World"
+	error := "404"
+	logContent := GetLogContent(logPrefix, pid, logLevel, filename, line, funcname, logMsg, error)
```
skdwriting: Why do we need to manually give pid, filename, line, funcname etc...? Please avoid that from the framework call. It can be automatically taken by the framework right? Please check how to optimize the same.
Why don't we use some standard go log packages and update as needed? (log, logrus, glog ..?)

pkg/utils/logs/logs.go
skdwriting: Why do we need to manually give pid, filename, line, funcname etc...? Please avoid that from the framework call. It can be automatically taken by the framework right? Please check how to optimize the same.
Why don't we use some standard go log packages and update as needed? (log, logrus, glog ..?)

pkg/utils/logs/logs.go
```diff
@@ -80,3 +81,24 @@ func InitLogs(LogFlushFrequency time.Duration) {
...
+	timeNow := time.Now().UTC().Format("2006-01-02 15:04:05.000000")
+	var logContent string
+	if logLevel != "ERROR" {
+		logContent = strings.Join([]string{"[", logLevel, "]", timeNow, logPrefix, "[ pid:", pid, "]", "[", fileName, ":", line, funcName, "]", logMsg}, " ")
```
skdwriting: Should we really need to do all these work? Let us try to use an existing log package and improve on it? Any feasibility issue?

pkg/utils/logs/logs.go
rhsakarpos: +1 to using a standard logger...i recommend using [https://github.com/uber-go/zap](url)

kumarashit: logrus is also one of the widely used package. It provides capability and flexibility which can be quite useful.

PR #988, opened 2019-08-21
Do not allow two profiles have same name

### LINKED ISSUE #987: Profile's name should be unique

/kind bug

**What happened**:
Profile should be unique,  so we should not have two profiles with same name.
Actually, we can not create two profiles with same name and same type, but we can **create two profiles with same name and different types**, which is unreasonable.

Can not create two profiles with same name, even though they have different types. 

Create a profile with name A and type B.
Then create another profile with name A and type C, this profile would be created successfully.

**Environment**:
- Hotpot(release/branch) version: latest
- OS (e.g. from /etc/os-release): Ubuntu 16.04.2 LTS (Xenial Xerus)
- Kernel (e.g. `uname -a`): Linux ubuntu 4.4.0-62-generic #83-Ubuntu SMP Wed Jan 18 14:10:15 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
- Install tools:
- Others:

ThisIsClark:
Profile's name should be unique, so check that whether the profile name is unique before creating a new profile.

**Which issue this PR fixes** :
 fixes #987

**Release note**:

```release-note
```

### DIFF
--- a/pkg/db/drivers/etcd/etcd.go
+++ b/pkg/db/drivers/etcd/etcd.go
@@ -1434,8 +1434,8 @@ func (c *Client) CreateProfile(ctx *c.Context, prf *model.ProfileSpec) (*model.P
 		prf.CreatedAt = time.Now().Format(constants.TimeFormat)
 	}
 
-	// profile name must be unique with the same storage type.
-	if _, err := c.getProfileByName(ctx, prf.Name, prf.StorageType); err == nil {
+	// profile name must be unique.
+	if _, err := c.getProfileByName(ctx, prf.Name); err == nil {
 		return nil, fmt.Errorf("the profile name '%s' already exists", prf.Name)
 	}
 
@@ -1476,7 +1476,23 @@ func (c *Client) GetProfile(ctx *c.Context, prfID string) (*model.ProfileSpec, e
 	return prf, nil
 }
 
-func (c *Client) getProfileByName(ctx *c.Context, name, storageType string) (*model.ProfileSpec, error) {
+func (c *Client) getProfileByName(ctx *c.Context, name string) (*model.ProfileSpec, error) {
+	profiles, err := c.ListProfiles(ctx)
+	if err != nil {
+		log.Error("List profile failed: ", err)
+		return nil, err
+	}
+
+	for _, profile := range profiles {
+		if profile.Name == name {
+			return profile, nil
+		}
+	}
+	var msg = fmt.Sprintf("can't find profile(name: %s)", name)
+	return nil, model.NewNotFoundError(msg)
+}
+
+func (c *Client) getProfileByNameAndType(ctx *c.Context, name, storageType string) (*model.ProfileSpec, error) {
 	profiles, err := c.ListProfiles(ctx)
 	if err != nil {
 		log.Error("List profile failed: ", err)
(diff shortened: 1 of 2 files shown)


pkg/db/drivers/etcd/etcd.go
```diff
@@ -1494,12 +1510,12 @@ func (c *Client) getProfileByName(ctx *c.Context, name, storageType string) (*mo
 
 // GetDefaultProfile
 func (c *Client) GetDefaultProfile(ctx *c.Context) (*model.ProfileSpec, error) {
-	return c.getProfileByName(ctx, "default", "block")
+	return c.getProfileByNameAndType(ctx, "default", "block")
```
wisererik: it's better to use constants instead of magic string

pkg/db/drivers/etcd/etcd.go
```diff
@@ -1494,12 +1510,12 @@ func (c *Client) getProfileByName(ctx *c.Context, name, storageType string) (*mo
...
 }
 
 // GetDefaultProfileFileShare
 func (c *Client) GetDefaultProfileFileShare(ctx *c.Context) (*model.ProfileSpec, error) {
-	return c.getProfileByName(ctx, "default", "file")
+	return c.getProfileByNameAndType(ctx, "default", "file")
 }
```
wisererik: Same here

pkg/db/drivers/etcd/etcd.go
kumarashit: Now the default profile created will be either default_block or default_file, so this call will always return nil. Right? Should we use default_block here

pkg/db/drivers/etcd/etcd.go
```diff
@@ -1494,12 +1510,12 @@ func (c *Client) getProfileByName(ctx *c.Context, name, storageType string) (*mo
...
 // GetDefaultProfile
 func (c *Client) GetDefaultProfile(ctx *c.Context) (*model.ProfileSpec, error) {
-	return c.getProfileByName(ctx, "default", "block")
+	return c.getProfileByNameAndType(ctx, "default", "block")
 }
 
 // GetDefaultProfileFileShare
 func (c *Client) GetDefaultProfileFileShare(ctx *c.Context) (*model.ProfileSpec, error) {
-	return c.getProfileByName(ctx, "default", "file")
+	return c.getProfileByNameAndType(ctx, "default", "file")
```
kumarashit: Same here as commented on 1497

pkg/db/drivers/etcd/etcd.go
ThisIsClark: I have changed the magic string to constant string.

pkg/db/drivers/etcd/etcd.go
ThisIsClark: Same as above

pkg/db/drivers/etcd/etcd.go
ThisIsClark: You are right.
I have changed this default name from "default" to "default_block"(same as the default name in install.sh)

pkg/db/drivers/etcd/etcd.go
ThisIsClark: Same as above

PR #998, opened 2019-08-29
Validation of file share mount paths for integration test

Shruthi-1MN:
Validation of file share mount paths for integration test
If file share already exists then only this file works fine
This test file is to ensure all the mount path exists on your operating system

If file share already exists then only this file works fine
This test file is to ensure all the mount path which are created through opensds exists on your operating system
It fails when no file share created from opensds exists on operating system and also when the path doesn't exists on operating system which are created from opensds. 

**Release note**:

```release-note
```


test/integration/paths_validation_test.go
```diff
@@ -0,0 +1,120 @@
...
+			panic(err)
+		}
+
+	if len(filesharelstmap) == 0{
+		t.Errorf("empty file share paths")
+	}
+	df := exec.Command("df")
+	grep := exec.Command("grep", "-w", "/mnt")
+	cut := exec.Command("awk", "{print $6}")//, "%", "-f", "1")
+
```
wisererik: remove unused code

test/integration/paths_validation_test.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func TrimSuffix(s, suffix string) string {
+	if strings.HasSuffix(s, suffix) {
+		s = s[:len(s)-len(suffix)]
+	}
+	return s
+}
+
+func contains(s []string, e string) bool {
```
kumarashit: Why not use contains()

PR #999, opened 2019-08-30
Integration test framework files for hotpot

satya-gorli:
The PR is for Integration test framework files using ginkgo
PR contains HTTPHelper (that connects via http and create http call), DBHelper ( which connects to dB and get data), LogReader ( which reads log file and do asseration)
IP of server need to be updated for OpenSDS hotpot installation path
**Release note**:
NONE
-->
```release-note
```

### DIFF
--- a/test/integration/fileshare_test.go
+++ b/test/integration/fileshare_test.go
@@ -1,0 +1,45 @@
+// Copyright 2019 The OpenSDS Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package integration
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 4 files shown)


test/integration/fileshare_test.go
```diff
@@ -0,0 +1,94 @@
...
+		})
+		It("TC_FS_IT_04: fileshare GET of specific Id", func() {
+			fId := "v1beta/file/shares/e93b4c0934da416eb9c8d120c5d04d96/f2ab9308-f208-40c6-bb1f-6fbfa8bf14b5"
+			url := "http://127.0.0.1.123:50040/v1beta/e93b4c0934da416eb9c8d120c5d04d96/file/shares/f2ab9308-f208-40c6-bb1f-6fbfa8bf14b5" //curl -X POST -H "Content-Type: application/json" -d '{"name":"share1", "description":"This is just for test", "size": 1}' -url "http://127.0.0.1:50040/v1beta/e93b4c0934da416eb9c8d120c5d04d96/file/shares"
+			methodName := "GET"
+			utils.ConnectToHTTP(methodName, url, nil)
+			// ctx, kv := utils.ConnectToDB()
+			ret := utils.GetValueByKeyFromDB(fId)
+			Expect(ret).ShouldNot(Equal(OPERATION_FAILED))
+			textFound := utils.ReadAndFindTextInFile("C:/go/src/opensds/opensds/test/integration/utils/output.json", "17c60641-63c9-4f7f-992a-c0dcd9abd502")
```
Shruthi-1MN: Which function writing this log file? Those entries should be written to this log file somewhere, where is that?

satya-gorli: this path is just for testing ... In actual testcase, what we want to find in log, need to give the log file path and text to find

Issue #997, opened 2019-08-28
New feature : Add wwn to volumespec


NajmudheenCT:
/kind feature

**What happened**:
Vmware Plugin code require wwn property in the LUN/volume. Vmware uses wwn to set disk id and path, through which the backing LUN/volume is identified.
Model change required
-------------------------
Model.volumespec should have wwn field. 
DB schema update required

Architecture change required 
--------------------------------
As of now the volume creation is an async call, we need to make it sync call so that the backend created wwn can be updated in reposnse.

driver change required
--------------------------
real backend ( storage arrays) will have wwn attached to LUN, driver can pick and fill volumespec.
Volume created by volume manager (lvm/ ceph?) can specify a unique id as scsi_id in the attachemnt and the same can be picked as wwn for the volume.

**What you expected to happen**:
volme POST/GET response should have wwn field.
**How to reproduce it (as minimally and precisely as possible)**:

wisererik:
It has an impact on driver, some of them doesn't have wwn, please discuss on community meeting

PR #1001, opened 2019-09-03
Implementation of Unique Durable Identifier for Volume

NajmudheenCT:
This is the implementation of model changes required for Unique Durable Identifier for Volume as mentioned in  https://github.com/opensds/design-specs/pull/47 .
This aslo includes the LVM driver implementaion.
As part of https://github.com/opensds/opensds/issues/997 , we need to add durable identifier filed in the volume to recognize backend LUN from Vmware platform. 

for  LVM, identifier  is the scsi_id created for iscsi tagets. In this PR we have changes in lvm driver to create scsi_id from volumeid . This scsi_id will be used to create targets and the same will be used as identifier. 

scsi_id is created from volID, whic is UUID generated dynamically.


contrib/drivers/lvm/lvm.go
```diff
@@ -48,8 +48,9 @@ const (
 )
 
 const (
-	KLvPath  = "lvPath"
-	KLvsPath = "lvsPath"
+	KLvPath     = "lvPath"
+	KLvsPath    = "lvsPath"
+	KLvIdFormat = "NAA"
```
kumarashit: Is this fine to keep it all uppercase? Vmware id is in all lowercase naa*

NajmudheenCT: Vmware append "naa" for their disk created form a NAA id. keeing "naa" in unique identifer is not required. As per swordfish durableidentifier is an enum , which has "NAA" in caps.
a sample response of swordfish API below. 
"Identifiers": [
{
"DurableNameFormat": "NAA",
"DurableName": "65456765456761001234076100123487"
}
],
We are also targeting the same format.

kumarashit: Ok

PravinRanjan10:
LGTM

PR #1011, opened 2019-09-12
Create Volume using Poolid

### LINKED ISSUE #978: New Feature: Create volume on given specific pool (poolid).

>
> /kind bug
> /kind feature

**What happened**:
My observation:
1. We need to create pool through opensds API or manual make entry in etcd db
2. Pool must be connected to some dock, so dock.id required.
3. We need to call getDockByPoolId() in createVolume method in controller block, to get dock info. 
4. Some small modification in controller code, then we can support this feature.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

NajmudheenCT:
To support volume and snapshot creation by specifiying poolid in request.
As of now ProfileID is mandatory to create volume. We enhance the support by adding volume creation by simply specifying pool .
fixes #978 
I have identified 3 usecase which is depending on Profile ID now.
1. While creating Volume
2. While creating snapshots
3. While creating volume from snapshot.

Please review and help me to identify if there is a new functionality which has got impacted or regressed due to this.


Shruthi-1MN:
It affects the back end storage capabilities, suppose if they specify profile and expecting to achieve the same by choosing the proper pool. If suppose the profile provided might match the default profile also, if it doesn't then it doesn't make sense.

NajmudheenCT:
> It affects the back end storage capabilities, suppose if they specify profile and expecting to achieve the same by choosing the proper pool. If suppose the doesn't provide profile, it should get pool by default profile. How are you choosing pool?

@Shruthi-1MN , As discussed  There is no option to pick default pool.  User should either have a default profile or specify a profile or spcify poolid. If poolid is specfied profile will be ignored, otherwise it will take defaut profile if exists.

pkg/api/controllers/volume.go
```diff
@@ -105,17 +116,26 @@ func (v *VolumePortal) CreateVolume() {
 		return
 	}
 	defer v.CtrClient.Close()
+	var profId string
+	var poolID string
+	if volume.ProfileId == "" {
```
PravinRanjan10: Wrong ProfileId will not be validated. If user provided wrong profileid, it should get validated and throw error message at API level only, before code flow goes to controller.

pkg/controller/controller.go
```diff
@@ -108,28 +108,47 @@ func (c *Controller) CreateVolume(contx context.Context, opt *pb.CreateVolumeOpt
...
+		//get poolname for poolid
+		pools, _ := db.C.ListPools(ctx)
+		for _, p := range pools {
+			if p.Id == opt.PoolId {
+				opt.PoolName = p.Name
+			}
+		}
+		// if poolname is still null set err
+		if opt.PoolName == "" {
+			err = errors.New(" No matching poolid found")
```
PravinRanjan10: Start the log message with lower case.

pkg/api/controllers/volume.go
NajmudheenCT: @PravinRanjan10 Here if condition block is when prfoilr id is not specified. incase of profileID is not cpecified the validation happens in Controller C.GetProfile() and the error reponse is like ""get profile failed: Wrong resource uuid provided!"

PravinRanjan10: As per discussed, please handle the case when poolid and wrong profileId given.

pkg/api/controllers/volume.go
```diff
@@ -66,21 +66,32 @@ func (v *VolumePortal) CreateVolume() {
...
-		prf, err = db.C.GetProfile(ctx, volume.ProfileId)
-	}
-	if err != nil {
-		errMsg := fmt.Sprintf("get profile failed: %s", err.Error())
-		v.ErrorHandle(model.ErrorBadRequest, errMsg)
-		return
+	// If pool id is not specified, get a profile or take default pool
+	if volume.PoolId == "" {
+		if volume.ProfileId == "" {
+			log.Warning("Use default profile when user doesn't specify profile.")
```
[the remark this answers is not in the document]
NajmudheenCT: Fixed thanks,

pkg/api/controllers/volume.go
```diff
@@ -66,21 +66,32 @@ func (v *VolumePortal) CreateVolume() {
...
+			prf, err = db.C.GetDefaultProfile(ctx)
+			// Assign the default profile id to volume so that users can know which
+			// profile is used for creating a volume.
+			if err == nil {
+				volume.ProfileId = prf.Id
+
+			} else {
+				//return Error: user should either have Profile or PoolID
+				errMsg := fmt.Sprintf("no pool id in the request or no default profile availble. %s", err.Error())
```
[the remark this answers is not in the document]
NajmudheenCT: fixed

pkg/api/controllers/volume.go
NajmudheenCT: @PravinRanjan10 , Thaks for the detailed review.
I have handled the case of having both profileid and pooid in request. also added validation ofr wrong poolid.

pkg/controller/controller.go
NajmudheenCT: fixed , thanks

pkg/api/controllers/volume.go
```diff
@@ -66,21 +66,45 @@ func (v *VolumePortal) CreateVolume() {
...
+		volume.ProfileId = ""
+		log.Warning("using only poolid when user provided both pool and profile .")
 	} else {
-		prf, err = db.C.GetProfile(ctx, volume.ProfileId)
-	}
-	if err != nil {
-		errMsg := fmt.Sprintf("get profile failed: %s", err.Error())
-		v.ErrorHandle(model.ErrorBadRequest, errMsg)
-		return
+		pool, err := db.C.GetPool(ctx, volume.PoolId)
```
PravinRanjan10: One doubt, If user doesn't add the pool details in DB, then GetPool will fail. Now next question is, how user will add the pool details into db?

NajmudheenCT: As discussed, As of now pool is a prerequiste for any operation in hotpot. 
pool creation is in the backend and hotpot only holds metadata of the pool. 
we can think of having create(POST ) /pool operation in hotpot.

pkg/controller/controller.go
```diff
@@ -108,28 +108,47 @@ func (c *Controller) CreateVolume(contx context.Context, opt *pb.CreateVolumeOpt
...
 	log.V(5).Infof("controller create volume:  get volume from db %+v", vol)
+	var dockInfo *model.DockSpec
+	if opt.ProfileId != "" || opt.SnapshotId != "" {
 
-	polInfo, err := c.selector.SelectSupportedPoolForVolume(vol)
-	if err != nil {
-		db.UpdateVolumeStatus(ctx, db.C, opt.Id, model.VolumeError)
-		return pb.GenericResponseError(err), err
-	}
+		polInfo, err := c.selector.SelectSupportedPoolForVolume(vol)
```
skdwriting: polnfo OR poolInfo ?

NajmudheenCT: changed to poolInfo

NajmudheenCT:
Closinf this PR as we dont want to create Volume without Profile and currently  If Default  Profile exists , providing poolid in the POST request will select the provided pool to create volume.

PR #1012, opened 2019-09-19
Bug Fix: For Volume Attachment to OpenSDS CSI plugin

himanshuvar:
nginx CSI example test is showing containerCreating state due to the docker not able to check the iscsi process state.
This PR will contains the changes required to check the iscsi process for the host not from the container.

**Which issue this PR fixes** 
fixes https://github.com/opensds/nbp/issues/248

### DIFF
--- a/contrib/connector/iscsi/helper.go
+++ b/contrib/connector/iscsi/helper.go
@@ -248,14 +248,18 @@ func connect(connMap map[string]interface{}) (string, error) {
 
 	targetlun := strconv.Itoa(conn.TgtLun)
 
-	cmd := "pgrep -f /sbin/iscsid"
-	_, err = connector.ExecCmd("/bin/bash", "-c", cmd)
+    cmd := "ls -ali / | sed '2!d' |awk {'print $1'}"
+    INODE_NUM, err := connector.ExecCmd("/bin/bash", "-c", cmd)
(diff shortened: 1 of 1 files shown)


contrib/connector/iscsi/helper.go
```diff
@@ -248,14 +248,18 @@ func connect(connMap map[string]interface{}) (string, error) {
 
 	targetlun := strconv.Itoa(conn.TgtLun)
 
-	cmd := "pgrep -f /sbin/iscsid"
-	_, err = connector.ExecCmd("/bin/bash", "-c", cmd)
+    cmd := "ls -ali / | sed '2!d' |awk {'print $1'}"
```
kumarashit: Any particular reason to use 'al'? All listing can be large at times and may not be efficient.

contrib/connector/iscsi/helper.go
```diff
@@ -248,14 +248,18 @@ func connect(connMap map[string]interface{}) (string, error) {
...
+    INODE_NUM, err := connector.ExecCmd("/bin/bash", "-c", cmd)
+    if err != nil && INODE_NUM == "2" {
+		cmd = "\"pgrep -f /sbin/iscsid\""
```
kumarashit: Suggestion: We can either get the daemon path from 'which' command and use it or define CONSTANT

PR #1022, opened 2019-10-11
Support pool status

### LINKED ISSUE #1015: osdsdock lvm backend gets duplicated entry on pool re-creation

/kind bug

**What happened**:

I noticed that osdsdock lvm driver can have duplicated entries 
when the underlying LVM VG was deleted once and created again.

I didn't see this this symptom for ceph driver.

Don't see duplicated entry.

1. Install opensds hotpot (at least ) and wait for stabilization.
  * You need to define at least 2 backends (e.g. lvm and ceph) in order to to avoid issue #991. (Or cherry-pick PR #1014.)
2. Delete opensds-volume VG by vgremove
3. Check 'osdsctl pool list' result. Defined 'opensds-volumes' pool does not dissapear.
4. Create opensds-volumes VG again by vgcreate.
5. You would see a new pool entry for the opensds-volumes like below. Old one you saw in step 3 is now orphan entry.

```
[ubuntu@ubuntu201 ~(opensds_admin)]$ date; osdsctl pool list
Fri Sep 20 09:38:32 UTC 2019
+--------------------------------------+-----------------+-------------+--------+---------------+--------------+
| Id                                   | Name            | Description | Status | TotalCapacity | FreeCapacity |
+--------------------------------------+-----------------+-------------+--------+---------------+--------------+
| f3ec5872-cd3b-51f0-86f3-0adc4778a77d | opensds-volumes |             |        | 10            | 10           |
 # Deleted (orphan) pool.
| 73efb059-1ce5-52fd-a379-01ced70851cf | osdsrbd         |             |        | 7             | 7            |
| 5f550e56-c613-502d-9eaa-97c31fb11bed | opensds-volumes |             |        | 10            | 10           |
 # New pool
+--------------------------------------+-----------------+-------------+--------+---------------+--------------+
[ubuntu@ubuntu201 ~(opensds_admin)]$
```

Maybe it's better to support 'osdsctl pool delete' command.

**Environment**:
- Hotpot(release/branch) version: capri
- OS (e.g. from /etc/os-release): Ubuntu 16.04.6
- Kernel (e.g. `uname -a`): Linux ubuntu201 4.4.0-157-generic #185-Ubuntu SMP Tue Jul 23 09:17:01 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
- Install tools: opensds-installer
- Others:

ThisIsClark:
When we query pool list, we can see a item "Status" which is always empty.
This PR add the status to each pool and it would be displayed in the pool list.
When the pool was deleted in backend, the status would be "unavailable", otherwise it would be "available".

**Which issue this PR fixes** : fixes #1015

**Special notes for your reviewer**: None

**Release note**:

```
root@ubuntu:~/gopath/src/github.com/opensds/opensds (master) #  ./build/out/bin/osdsctl pool list
WARNING: OPENSDS_ENDPOINT is not specified, use default(http://localhost:50040)
WARNING: Not found Env OPENSDS_AUTH_STRATEGY, use default(noauth)
+--------------------------------------+-------------------------+-------------+-----------+---------------+--------------+
| Id                                   | Name                    | Description | Status    | TotalCapacity | FreeCapacity |
+--------------------------------------+-------------------------+-------------+-----------+---------------+--------------+
| 18272d49-3c58-59f2-82a5-0a5134786ba2 | opensds-volumes-default |             | available | 20            | 20           |
| 16e17837-0319-5381-bd67-0cbfe73d3ff7 | opensds-files-default   |             | available | 20            | 20           |
+--------------------------------------+-------------------------+-------------+-----------+---------------+--------------+
root@ubuntu:~/gopath/src/github.com/opensds/opensds (master) # vgremove opensds-volumes-default
  Volume group "opensds-volumes-default" successfully removed
root@ubuntu:~/gopath/src/github.com/opensds/opensds (master) # ./build/out/bin/osdsctl pool list
WARNING: OPENSDS_ENDPOINT is not specified, use default(http://localhost:50040)
WARNING: Not found Env OPENSDS_AUTH_STRATEGY, use default(noauth)
+--------------------------------------+-------------------------+-------------+-------------+---------------+--------------+
| Id                                   | Name                    | Description | Status      | TotalCapacity | FreeCapacity |
+--------------------------------------+-------------------------+-------------+-------------+---------------+--------------+
| 18272d49-3c58-59f2-82a5-0a5134786ba2 | opensds-volumes-default |             | unavailable | 20            | 20           |
| 16e17837-0319-5381-bd67-0cbfe73d3ff7 | opensds-files-default   |             | available   | 20            | 20           |
+--------------------------------------+-------------------------+-------------+-------------+---------------+--------------+
root@ubuntu:~/gopath/src/github.com/opensds/opensds (master) #
```

### DIFF
--- a/pkg/dock/discovery/discovery.go
+++ b/pkg/dock/discovery/discovery.go
@@ -186,7 +186,7 @@ func (pdd *provisionDockDiscoverer) Discover() error {
 	ctx := c.NewAdminContext()
 	polsInDb, err = pdd.c.ListPools(ctx)
 	if err != nil {
-		return fmt.Errorf("Can not read pools in db")
+		return fmt.Errorf("can not read pools in db")
 	}
 	dbPolsMap := make(map[string]*model.StoragePoolSpec)
 	for _, polInDb := range polsInDb {
@@ -200,7 +200,7 @@ func (pdd *provisionDockDiscoverer) Discover() error {
 		pdd.pols = append(pdd.pols, unavailablePol)
 	}
 	if len(pdd.pols) == 0 {
-		return fmt.Errorf("There is no pool can be found.")
+		return fmt.Errorf("there is no pool can be found")
 	}
 
 	return nil
--- a/pkg/dock/discovery/discovery.go
+++ b/pkg/dock/discovery/discovery.go
@@ -39,6 +39,11 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
+const (
+	availableStatus = "available"
+	unavailableStatus = "unavailable"
+)
+
 type Context struct {
 	StopChan chan bool
 	ErrChan  chan error
@@ -135,6 +140,7 @@ func (pdd *provisionDockDiscoverer) Discover() error {
 	// Clear existing pool info
 	pdd.pols = pdd.pols[:0]
 	var pols []*model.StoragePoolSpec
+	var polsInDb []*model.StoragePoolSpec
 	var err error
 	for _, dck := range pdd.dcks {
 		// Call function of StorageDrivers configured by storage drivers.
@@ -145,6 +151,7 @@ func (pdd *provisionDockDiscoverer) Discover() error {
 			for _, pol := range pols {
 				log.Infof("Backend %s discovered pool %s", dck.DriverName, pol.Name)
 				pol.DockId = dck.Id
+				pol.Status = availableStatus
 			}
 		} else {
 			d := drivers.Init(dck.DriverName)
@@ -162,6 +169,7 @@ func (pdd *provisionDockDiscoverer) Discover() error {
 				pol.DockId = dck.Id
 				pol.ReplicationType = replicationType
 				pol.ReplicationDriverName = replicationDriverName
+				pol.Status = availableStatus
 			}
 		}
 		if err != nil {
@@ -174,6 +182,22 @@ func (pdd *provisionDockDiscoverer) Discover() error {
 		}
 
 		pdd.pols = append(pdd.pols, pols...)
+	}
+	ctx := c.NewAdminContext()
+	polsInDb, err = pdd.c.ListPools(ctx)
+	if err != nil {
+		return fmt.Errorf("Can not read pools in db")
+	}
+	dbPolsMap := make(map[string]*model.StoragePoolSpec)
+	for _, polInDb := range polsInDb {
+		dbPolsMap[polInDb.Id] = polInDb
+	}
+	for _, pol := range pdd.pols {
+		delete(dbPolsMap, pol.Id)
+	}
+	for _, unavailablePol := range dbPolsMap {
+		unavailablePol.Status = unavailableStatus
+		pdd.pols = append(pdd.pols, unavailablePol)
 	}
 	if len(pdd.pols) == 0 {
 		return fmt.Errorf("There is no pool can be found.")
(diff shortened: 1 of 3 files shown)


ThisIsClark:
Hi @kumarashit 
> Generic query: we are making the status hardcoded. i.e. not actually issuing any API/CLI to query the current status. Is this fine?

Sorry, it seems I didn't fully get your point. Could you please explain it more detailly?

> Also, status can be in-error, busy or something different from available and unavailable???

Nowadays I only set available and unavailable according to whether the pool is actually exist in the backend.
And we can discuss about what other values could be add to key "status" besides "available" and "unavailable".
> Doubt: If the pool is deleted, why it appears in the list?

If we simply delete the pools info from opensds when the pool was deleted, we need to handle all the other info in opensds witch is related to this deleted pool, such as volume info.
So I think maybe keep this pool info but set its status to "unavailable" is a better option.
And for these redundancy info(which status is "unavailable"),  we should allow users to delete them manually.

kumarashit:
> 
> 
> Hi @kumarashit
> 
> > Generic query: we are making the status hardcoded. i.e. not actually issuing any API/CLI to query the current status. Is this fine?
> 
> Sorry, it seems I didn't fully get your point. Could you please explain it more detailly?
> From what i  get from the code, we are actually not running any pool status API/CLI. i.e., if we get the list of pools, we are setting it as available. It maybe that the pool i got is in error state and we have marked as available.
> > Also, status can be in-error, busy or something different from available and unavailable???
> 
> Nowadays I only set available and unavailable according to whether the pool is actually exist in the backend.
> And we can discuss about what other values could be add to key "status" besides "available" and "unavailable".
> Sure
> > Doubt: If the pool is deleted, why it appears in the list?
> 
> If we simply delete the pools info from opensds when the pool was deleted, we need to handle all the other info in opensds witch is related to this deleted pool, such as volume info.
>> Right and that should be the correct way. Isn't it?
> So I think maybe keep this pool info but set its status to "unavailable" is a better option.
> And for these redundancy info(which status is "unavailable"), we should allow users to delete them manually.

pkg/dock/discovery/discovery.go
```diff
@@ -175,6 +183,22 @@ func (pdd *provisionDockDiscoverer) Discover() error {
 
 		pdd.pols = append(pdd.pols, pols...)
 	}
+	ctx := c.NewAdminContext()
+	polsInDb, err = pdd.c.ListPools(ctx)
+	if err != nil {
+		return fmt.Errorf("Can not read pools in db")
```
PravinRanjan10: Start the message with lower letter.

Other thing looks ok to me

ThisIsClark: Done

ThisIsClark:
Hi @kumarashit ,
I have thought about your comment again, I think this may be not feasible:

> b) Check Pool status from the native commands

Because we can get status from vgdisplay for lvm, but not every the backend has the command to get the pool's status, so maybe keeping these two status "available" and "unavailable" is a better choice.

kumarashit:
> Hi @kumarashit ,
> I have thought about your comment again, I think this may be not feasible:
> 
> > b) Check Pool status from the native commands
> 
> Because we can get status from vgdisplay for lvm, but not every the backend has the command to get the pool's status, so maybe keeping these two status "available" and "unavailable" is a better choice.



> Hi @kumarashit ,
> I have thought about your comment again, I think this may be not feasible:
> 
> > b) Check Pool status from the native commands
> 
> Because we can get status from vgdisplay for lvm, but not every the backend has the command to get the pool's status, so maybe keeping these two status "available" and "unavailable" is a better choice.

Ok. 
Did you check? Because any backend resource group or pool should provide this info.

PR #1091, opened 2019-11-28
Use known version of etcd

### LINKED ISSUE #1090: Installer script in devsds does not work if etcd is already installed with a different version
OpenSDS devsds installer do not consider version of etcd if etcd is already installed.

/kind bug

**What happened**:
When trying to install opensds using 'install/devsds/install.sh', in a system with already installed etcd, installer failed.

Installer should be using a consistent version of etcd as configured.

Install latest etcd 3.4 in the system and try to run installer script.

joseph-v:
This PR forces devsds installer to use a known version of etcd

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1090


install/devsds/lib/etcd.sh
```diff
@@ -41,9 +41,9 @@ osds::etcd::download() {
 
 osds::etcd::install() {
     # validate before running
-    which etcd >/dev/null || {
-    osds::etcd::download
-    }
+    if [ ! -f "${OPT_DIR}/etcd-v${ETCD_VERSION}-linux-amd64/etcd" ]; then
+        osds::etcd::download
+    fi
```
wisererik: it's alse be fine if etcd is install on the system path

joseph-v: A etcd latest version (3.4) will not work with installer if it is present in system path. We need etcd version 3.3

wisererik: ok

PR #1097, opened 2019-12-04
Fix for fileshare mount issue

PravinRanjan10:

Corrected the returing export location. Previous it was, [IP:/var/<fs name>]

now it is: [IP:/mnt/<fs name>]

**Release note**:

```release-note
```

### DIFF
--- a/contrib/drivers/filesharedrivers/nfs/cli.go
+++ b/contrib/drivers/filesharedrivers/nfs/cli.go
@@ -77,8 +77,8 @@ func (c *Cli) CreateAccess(accessto, accesscapability, fname string) error {
 
 func (c *Cli) DeleteAccess(accessto, fname string) error {
 	var accesstoAndMount string
-	sharePath := path.Join("mnt/", fname)
-	accesstoAndMount = fmt.Sprintf("%s:/%s", accessto, strings.Replace(sharePath, "-", "_", -1))
+	sharePath := path.Join(MountPath, fname)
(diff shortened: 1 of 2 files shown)


PravinRanjan10:
fs1
This is local test report

wisererik:
it's confused that changing /var/ to /mnt/, can you explaint it here?

PravinRanjan10:
> it's confused that changing /var/ to /mnt/, can you explaint it here?

Yes. Actually fileshare is mount at /mnt path only. so export location should be at /mnt path only. Previously it was wrongly hardcoded to /var path. That is the reason i have changed to /mnt

wisererik:
> > it's confused that changing /var/ to /mnt/, can you explaint it here?
> 
> Yes. Actually fileshare is mounted at /mnt path only. so export location should be at /mnt path only. Previously it was wrongly hardcoded to /var path. That is the reason i have changed to /mnt



> > it's confused that changing /var/ to /mnt/, can you explaint it here?
> 
> Yes. Actually fileshare is mounted at /mnt path only. so export location should be at /mnt path only. Previously it was wrongly hardcoded to /var path. That is the reason i have changed to /mnt

it's weird that it's not permitted if user wants to share /home/username/sharepath. is there any document mention this "fileshare is mounted at /mnt path only"?

wisererik:
please use constant to replace magic string

PravinRanjan10:
> please use constant to replace magic string

ok

PravinRanjan10:
resolved all comments. Please review and merge

PR #1194, opened 2020-01-21
Support multipath device

sushanthakumar:
This PR is for supporting multipath device

This PR fixes the issue : https://github.com/opensds/opensds/issues/1196

> Any other PR(s) this PR is dependant on: None

Test steps:
Install multipath package at host
Create multiple mounts through creation of multiple pods
Verify the mount paths as host

**Release note**:

```release-note
```

### DIFF
--- a/contrib/connector/fc/fibreChannel.go
+++ b/contrib/connector/fc/fibreChannel.go
@@ -81,6 +81,14 @@ func connectVolume(connMap map[string]interface{}) (map[string]string, error) {
 		return nil, err
 	}
 
+	dmPath, err := getMultipathDevice(deviceWWN)
+	if err != nil {
+		return nil, err
+	}
+	if len(dmPath) > 0 {
+		devicePath = dmPath
+	}
+
 	return map[string]string{"scsi_wwn": deviceWWN, "path": devicePath}, nil
 }
 
--- a/contrib/connector/fc/linuxfc.go
+++ b/contrib/connector/fc/linuxfc.go
@@ -181,9 +181,9 @@ func getMultipathDevice(deviceWWN string) (string, error) {
 	cmd := fmt.Sprintf("ls -l /dev/disk/by-id/ | grep %s", deviceWWN)
 	out, err := connector.ExecCmd("/bin/bash", "-c", cmd)
 	if err != nil {
-		errMsg := fmt.Sprintf("error occurred when find DM of wwn %s: %s, %v", deviceWWN, out, err)
-		log.Println(errMsg)
-		return "", errors.New(errMsg)
+		msg := fmt.Sprintf("No DM of wwn %s exist", deviceWWN)
+		log.Println(msg)
+		return "", nil
 	}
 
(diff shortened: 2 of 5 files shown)


contrib/connector/fc/linuxfc.go
```diff
@@ -176,3 +176,28 @@ func trimDoubleQuotesInText(str string) string {
 	}
 	return str
 }
+
+func getMultipathDevice(deviceWWN string) (string, error) {
```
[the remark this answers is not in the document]
kumarashit: Also, do we check somewhere if multipath is enabled on the host or not? Will it be a better idea to check if /etc/multipath conf exists and then if yes, check if the device has multipath

contrib/connector/fc/fibreChannel.go
```diff
@@ -81,6 +81,14 @@ func connectVolume(connMap map[string]interface{}) (map[string]string, error) {
 		return nil, err
 	}
 
+	dmPath, err := getMultipathDevice(deviceWWN)
```
NajmudheenCT: Please help me to understand the multipath use case here..
Which API of OpenSDS depends on this functionality ? Volume Attach? 
 Why do we need to collect information's from the host to attach? Do we support attachment only to the host where dock is running?

contrib/connector/fc/linuxfc.go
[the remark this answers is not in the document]
zengyingzhe: this code doesn't care if multipath enabled or not on host, it tries to get multipath device first, if not exist, then tries to find normal sd device instead.
But I alse think it would better add some kind of configuration item to allow user to specify whether or not using mulitipath.
I'll think about it.

contrib/connector/fc/fibreChannel.go
zengyingzhe: yes, volume attach involves this multipath function. if there are multiple paths connecting between host and storage, we must use multipath device instead of normal sd device to utilize the multipath HA.

zengyingzhe: multipath device is a host dev file, like /dev/dm-*, we need to find out this file path and use it.

zengyingzhe: we can attach volume to any host that opensds aware, not only for dock host.

contrib/connector/fc/linuxfc.go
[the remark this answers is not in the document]
sushanthakumar: True, we can give some config for enable/disabling this feature, we can open an issue for this and think about how we can do, thanks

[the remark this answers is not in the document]
kumarashit: Ok. So this code can be additional check. We can go with it. For multipath enabled/disabled, i think Host already provides the /etc/multipath file if multipath is enabled or not. we can use that.

contrib/connector/fc/fibreChannel.go
NajmudheenCT: ok

PR #1195, opened 2020-01-22
[Driver] Enhancement for concurrency

sushanthakumar:
This PR is for enhancement for concurrency scenarios at driver side

**Release note**:

```release-note
```

### DIFF
--- a/contrib/drivers/drivers.go
+++ b/contrib/drivers/drivers.go
@@ -85,9 +85,14 @@ type VolumeDriver interface {
 	ListPools() ([]*model.StoragePoolSpec, error)
 }
 
+var (
+	OceanStorDriver = oceanstor.Driver{}
+)
+
 // Init
-func Init(resourceType string) VolumeDriver {
+func Init(resourceType string) (VolumeDriver, error) {
 	var d VolumeDriver
+
 	switch resourceType {
 	case config.CinderDriverType:
 		d = &cinder.Driver{}
@@ -102,7 +107,7 @@ func Init(resourceType string) VolumeDriver {
 		d = &spectrumscale.Driver{}
 		break
 	case config.HuaweiOceanStorBlockDriverType:
-		d = &oceanstor.Driver{}
+		d = &OceanStorDriver
 		break
 	case config.HuaweiFusionStorageDriverType:
 		d = &fusionstorage.Driver{}
@@ -119,12 +124,17 @@ func Init(resourceType string) VolumeDriver {
 		d = &sample.Driver{}
 		break
 	}
-	d.Setup()
-	return d
+
+	err := d.Setup()
+	if err != nil {
+		return nil, err
+	}
+
+	return d, nil
 }
 
 // Clean
-func Clean(d VolumeDriver) VolumeDriver {
+func Clean(d VolumeDriver) {
 	// Execute different clean operations according to the VolumeDriver type.
 	switch d.(type) {
 	case *cinder.Driver:
@@ -136,7 +146,7 @@ func Clean(d VolumeDriver) VolumeDriver {
 	case *spectrumscale.Driver:
 		break
 	case *oceanstor.Driver:
-		break
+		return // No need to clean anything for oceanstor
 	case *fusionstorage.Driver:
 		break
 	case *nimble.Driver:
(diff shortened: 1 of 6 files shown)


pkg/dock/dock.go
```diff
@@ -419,16 +464,21 @@ func (ds *dockServer) UpdateVolumeGroup(ctx context.Context, opt *pb.UpdateVolum
 
 func (ds *dockServer) DeleteVolumeGroup(ctx context.Context, opt *pb.DeleteVolumeGroupOpts) (*pb.GenericResponse, error) {
 	// Get the storage drivers and do some initializations.
-	ds.Driver = drivers.Init(opt.GetDriverName())
-	defer drivers.Clean(ds.Driver)
+	driver, err := drivers.Init(opt.GetDriverName())
```
kumarashit: Why cannot be this part of the code moved to a new func, say init() and called everywhere? It is repetition of code

kumarashit:
The issue of concurrency happens just because there is no proper error handling?
Can there be a case that:
a) Volume is created and attach request is made
b) Port X, is specified but the Port status is not changed from available->unavailable yet
c) Another Volume created and attach request made
d) It finds that Port X is still available, now there may be a race condtion
e) Should we wait for the next attachment request or keep a list of available and unavailable port? May be some network lag may cause attach/detach request to fall into race condition.
Just thinking...

contrib/drivers/drivers.go
```diff
@@ -136,7 +146,7 @@ func Clean(d VolumeDriver) VolumeDriver {
 	case *spectrumscale.Driver:
 		break
 	case *oceanstor.Driver:
-		break
+		return // No need to clean anything for oceanstor
```
PravinRanjan10: What happen, if i really mean to unset/clean the driver??

contrib/drivers/drivers.go
```diff
@@ -148,10 +158,9 @@ func Clean(d VolumeDriver) VolumeDriver {
 	default:
 		break
 	}
+
 	d.Unset()
```
PravinRanjan10: It's better to handle Unset error also. I have seen you have handled the error in Setup()

pkg/dock/dock.go
zengyingzhe: I just followed the original code logic, however, yes it is repeated.
I'd rather leave this improvement to another pr.

contrib/drivers/drivers.go
```diff
@@ -102,7 +107,7 @@ func Init(resourceType string) VolumeDriver {
 		d = &spectrumscale.Driver{}
 		break
 	case config.HuaweiOceanStorBlockDriverType:
-		d = &oceanstor.Driver{}
+		d = &OceanStorDriver
```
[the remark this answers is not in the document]
zengyingzhe: OceanStorDriver is defined as a global variable, so each time we call Init for Huawei OceanStor driver, we can get the same driver object, not create a new one as before.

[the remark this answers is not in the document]
zengyingzhe: This's also for the concurrency circumstance, because each oceanstor driver maintains a connect to OceanStor storage, if there are big concurrency invokes, the connection number may exceed the limit of OceanStor.

contrib/drivers/drivers.go
zengyingzhe: oceanstor driver's unset operation just close the connection to back storage, cause we use the same global oceanstor driver for each call now, so there's no need to unset anymore. Instead, we want to keep the connection session active consistently.

contrib/drivers/drivers.go
zengyingzhe: as the old code's implementation, I think the unset error is ignorable, so I didn't change here.

pkg/dock/dock.go
kumarashit: Ok.

kumarashit:
Can you please check CI issues

kumarashit:
@sushanthakumar Please check the CI issue

contrib/drivers/drivers.go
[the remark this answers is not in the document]
PravinRanjan10: ok

contrib/drivers/drivers.go
PravinRanjan10: ok

contrib/drivers/drivers.go
PravinRanjan10: ok

PR #1299, opened 2020-06-17
Unit test cases for block volume and block volume snapshot

nguptaopensds:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind new feature
> /kind bug fix
> /kind cleanup
> /kind revert change
> /kind design
> /kind documentation

 /kind enhancement

Adding unit test cases for block volumes

**Which issue(s) this PR fixes**:
#1270 

Fixes #

**Test Report Added?**:
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:

 /kind TESTED
> /kind NOT-TESTED

**Test Report**:


pkg/api/controllers/volume_test.go
```diff
@@ -96,6 +98,64 @@ var (
...
+			httpCtx.Input.SetData("context", c.NewAdminContext())
+		})
+		beego.BeeApp.Handlers.ServeHTTP(w, r)
+		var output model.VolumeSpec
+		json.Unmarshal(w.Body.Bytes(), &output)
+		assertTestResult(t, w.Code, 202)
+		assertTestResult(t, &output, &SampleVolumes[0])
+	})
+
+	t.Run("Should return 500 if create volume with bad request", func(t *testing.T) {
```
kumarashit, line 141: why 500

testutils/collection/data.go
```diff
@@ -375,6 +375,18 @@ var (
 			ProfileId:        "1106b972-66ef-11e7-b172-db03f3689c9c",
 			SnapshotId:       "3769855c-a102-11e7-b772-17b880d2f537",
 		},
+		{
```
kumarashit, line 378: can existing be re-used?

nguptaopensds: When we run the test for create then the status of the structure changes to "creating", same with delete it changes to "deleting", so rather than using the same struct across we can use a copy for those 2 different test cases, else when we run test on all test cases together it fails, saying status should be availble not creating/deleting

pkg/api/controllers/volume_test.go
nguptaopensds: I will again update and check for bad request , may be we need to change main code as well

kumarashit: Is this resolved?

nguptaopensds: Resolved

kumarashit (approved):
LGTM

